PR #248, opened 2024-07-02
Adding Husky To Manage Git Pre Commit Hooks

splindsay-92:
Adding husky and a pre commit hook to run prettier and linter.


splindsay-92:
I think this is worth re-visiting at a later date, so I will close the PR.

PR #249, opened 2024-07-04
feature: Added Public Ably Message Parsers 

splindsay-92:
### Context

* [CHA-337]
* Users may wish to make use of ably's integration offerings, such as webhooks/AWS lambdas, where messages/reactions etc. sent to the room can be filtered and sent to an integration. It would be useful to have some functions set up that users can pass an encoded inbound message-like object and decode them straight into a chat type like a message or reaction.

### Description

- Added two functions that take an encoded ably message, such as those sent via integration rules, and try to convert them into chat types. 
  - Added fromEncodedChatMessage function that will try and convert the message into a chat message. 
  - Added fromEncodedReaction function that will try and convert the message into a reaction.

- Added getEntityTypeFromEncoded function that will also take an encoded message, and attempt to resolve the chat entity type, currently either a chat message or reaction. This can then be used to inform the user which of the two fromEncoded function you need to use.

- Added two functions that take an ably inbound message, and try to convert them into chat types. 
  - Added fromAblyMessageToReaction function that will take an ably inboundMessage and try to convert it to a reaction
  - Added fromAblyMessageToChatMessage function that will take an ably inboundMessage and try to convert it to a chat message.

- Added getEntityTypeFromAblyMessage function that will take inboundMessage, and attempt to resolve the chat entity type, currently either a chat message or reaction. This can then be used to inform the user which of the two fromAblyMessage functions you need to use.

- Moved message and reaction parsing to internal folders for reuse outside their previous classes. Added testing for message and reaction parsing since it wasn't covered comprehensively.

### Checklist

* [x] QA'd by the author.

[CHA-337]: https://ably.atlassian.net/browse/CHA-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,6 +567,104 @@ unsubscribe();
 
 You can remove all listeners at once by calling `reactions.unsubscribeAll()`.
 
+## Handling encoded objects in integrations
+
+If you have set up an [Ably integration](https://ably.com/docs/general/integrations) to receive events from your chat
+room, depending on your configuration, you may receive these as encoded objects.
+See [here](https://ably.com/docs/general/webhooks) for more information.
+Should you wish to convert this object to a chat type, you can use the functions below to help you.
+
+For example, if you have the following item reach your integration:
+
+```json
+{
+  "items": [
+    {
+      "webhookId": "Ja-tsg",
+      "source": "channel.message",
+      "serial": "108iZpUxQBe4Vv35120919@1720954404104-0",
+      "timestamp": 1720954404104,
+      "name": "channel.message",
+      "data": {
+        "channelId": "some-room::$chat::$chatMessages",
+        "site": "eu-west-1-A",
+        "messages": [
+          {
+            "id": "chat:6TP2sA:some-room:a4534b0ab37bdd5:0",
+            "clientId": "user1",
+            "timestamp": 1720954404104,
+            "encoding": "json",
+            "extras": {
+              "timeserial": "108iZpUxQBe4Vv35120919@1720954404104-0",
+              "headers": {}
+            },
+            "data": "{\"text\":\"some text data\",\"metadata\":{}}",
+            "name": "message.created"
+          }
+        ]
+      }
+    }
+  ]
+}
+```
+
+You should use `getEntityTypeFromEncoded` to first retrieve the chat entity type of the encoded message,
+then call either `chatMessageFromEncoded` or `reactionFromEncoded` depending on the entity type.
+
+```ts
+import { getEntityTypeFromEncoded, chatMessageFromEncoded, reactionFromEncoded, ChatEntityType } from '@ably/chat';
+
+integrationMessage.items.forEach((item) => {
+  item.data.messages.forEach(async (encodedMessage) => {
+    const entityType = getEntityTypeFromEncoded(encodedMessage);
+    switch (entityType) {
+      case ChatEntityType.ChatMessage:
+        const chatMessage = await chatMessageFromEncoded(encodedMessage);
+        console.log(chatMessage);
+        break;
+      case ChatEntityType.Reaction:
+        const reaction = await reactionFromEncoded(encodedMessage);
+        console.log(reaction);
+        break;
+      default:
+        console.log('Unknown entity type');
+    }
+  });
+});
+```
+
+## Handling Ably messages
+
+If you are working with the underlying channels directly and not using the Chat SDK, you can use these functions to
+convert an inbound Ably message to a chat type.
+You can use `getEntityTypeFromAblyMessage` to retrieve the chat entity type of the message,
+then call either `chatMessageFromAblyMessage` or `reactionFromAblyMessage` depending on the entity type.
+
+```ts
+import * as Ably from 'ably';
+import {
+  getEntityTypeFromAblyMessage,
+  chatMessageFromAblyMessage,
+  reactionFromAblyMessage,
+  ChatEntityType
+} from '@ably/chat';
+
+const entityType = getEntityTypeFromAblyMessage(inboundMessage as Ably.InboundMessage);
+
+switch (entityType) {
+  case ChatEntityType.ChatMessage:
+    const chatMessage = chatMessageFromAblyMessage(inboundMessage as Ably.InboundMessage);
+    console.log(chatMessage);
+    break;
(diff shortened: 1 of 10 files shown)


src/internal/parsers/messageParser.ts
```diff
@@ -0,0 +1,57 @@
+import { InboundMessage } from 'ably';
```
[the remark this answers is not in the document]
splindsay-92: I've moved this out into root level for now to avoid a major refactor, might be worth doing this as the codebase grows though.

README.md
```diff
@@ -567,6 +567,105 @@ unsubscribe();
...
+            "name": "message.created"
+          }
+        ]
+      }
+    }
+  ]
+}
+```
+
+If you are unsure as to the underlying chat entity type of the encoded message,
```
AndyTWF, line 611: Should this be the default recommendation, that you should use `getEntityTypeFromEncoded`?

README.md
```diff
@@ -567,6 +567,105 @@ unsubscribe();
...
+        console.log(reaction);
+        break;
+      default:
+        console.log('Unknown entity type');
+    }
+  });
+});
+```
+
+## Handling ably inbound messages
```
AndyTWF, line 637:
```suggestion
## Handling Ably inbound messages
```

README.md
```diff
@@ -567,6 +567,105 @@ unsubscribe();
...
+
+It may be the case that you wish to convert an Ably inboundMessage to a chat type, such as if you are working with the
```
AndyTWF, line 639: We don't really document `InboundMessage` anywhere as a thing - it's niche to ably-js. Perhaps something like "If you're working with Channels directly and not using the Chat SDK?"

src/Messages.ts
```diff
@@ -503,68 +504,12 @@ export class DefaultMessages
...
+      return parseMessage(this._roomId, channelEventMessage);
     } catch (error: unknown) {
-      this._logger.error(`failed to parse incoming message`, {
-        channelEventMessage,
-        error: error as Ably.ErrorInfo,
-      });
-      return;
+      this._logger.error(`failed to parse incoming message;`, { channelEventMessage, error: error as Ably.ErrorInfo });
     }
+    return;
```
AndyTWF, line 512: `return` not needed here - not returning anything will just yield `undefined`

src/RoomReactions.ts
```diff
@@ -248,38 +249,13 @@ export class DefaultRoomReactions
...
-
-    return new DefaultReaction(
-      data.type,
-      inbound.clientId,
-      new Date(inbound.timestamp),
-      inbound.clientId === clientId,
-      data.metadata ?? {},
-      extras?.headers ?? {},
-    );
+    return;
```
AndyTWF, line 258: Ditto with messages

src/helpers.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+ * Helper function to validate the encoded object.
+ *
+ * @param encoded - The object to validate.
+ * @throws {@link ErrorInfo} - If the encoded object is invalid.
+ */
+function validateEncoded(encoded: unknown): void {
+  if (typeof encoded !== 'object' || encoded === null) {
```
AndyTWF, line 31: Do we need the `encoded === null` part here? Won't this case just be caught by the first condition?

src/messageParser.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { DefaultMessage, Message, MessageHeaders, MessageMetadata } from './Message.js';
+
+export interface MessagePayload {
+  data?: {
+    text?: string;
+    metadata?: MessageMetadata;
+  };
+  clientId?: string;
+  timestamp?: number;
```
AndyTWF, line 11: `timestamp` cannot be optional in an `InboundMessage`

src/messageParser.ts
```diff
@@ -0,0 +1,57 @@
+import * as Ably from 'ably';
+
+import { DefaultMessage, Message, MessageHeaders, MessageMetadata } from './Message.js';
+
+export interface MessagePayload {
```
AndyTWF, line 5: Are we using this outside this file? If not, we don't need to export it.

src/reactionParser.ts
```diff
@@ -0,0 +1,43 @@
+import * as Ably from 'ably';
+
+import { DefaultReaction, Reaction, ReactionHeaders, ReactionMetadata } from './Reaction.js';
+
+export interface ReactionPayload {
```
AndyTWF, line 5: Ditto re export

src/reactionParser.ts
```diff
@@ -0,0 +1,43 @@
...
+  data?: {
+    type: string;
+    metadata?: ReactionMetadata;
+  };
+  clientId?: string;
+  timestamp?: number;
```
AndyTWF, line 11: Ditto re timestamp

src/helpers.ts
splindsay-92: Sorry, I don't quite understand, probably my ts knowledge is lacking here, but the `typeof null === 'object'` is true, so doesn't that mean encoded can be null and pass the check if we don't check specifically for null? :)

AndyTWF: Well oh my... JavaScript never fails to surprise. Ignore this!

src/messageParser.ts
splindsay-92: No, I just missed it, will remove the export

README.md
splindsay-92: Yeh I agree with you. I had thought it would depend on how you set up your integration, i.e. if you choose only to send reactions to your integration via channel rule filter, but by setting the default recommendation as always calling `getEntityTypeFromEncoded` you catch that case anyway :)

README.md
splindsay-92: I can change the blurb here and the title above, but considering the types are inbound messages, it feels a bit strange to obscure that fact here in the readme?

splindsay-92: Ahh I see, yes in that case I'll update this to your phrasing there.

AndyTWF (approved):

PR #269, opened 2024-07-09
Update README.md

Srushtika:


AndyTWF:
> I'm putting my current review comments on the README in here.
> 
> https://github.com/ably/ably-chat-js/blob/d65ada88ed2d3b06da3cb7daf11c96b6948f3669/README.md#installation-and-authentication - why are these two topics under the same heading?
> 
> https://github.com/ably/ably-chat-js/blob/d65ada88ed2d3b06da3cb7daf11c96b6948f3669/README.md#current-connection-status: should this and the following headings be h3s? I think they should be subordinate to the "Connections" heading.
> 
> https://github.com/ably/ably-chat-js/blob/d65ada88ed2d3b06da3cb7daf11c96b6948f3669/README.md#creating-or-retrieving-a-chat-room: "Sensible defaults are provided ...." - could there be a link here to say what options there are, and what the defaults are?
> 
> "Headers are a flat key-value map and are sent as part of the realtime message's extras inside the headers property" - we need backticks around `extras`, `headers` etc.
> 
> There are some sections that we normally include in a README that are missing, eg:
> 
> * a link to contributing (eg https://github.com/ably/ably-js/blob/main/README.md#contributing)
> * compatibility (eg https://github.com/ably/ably-js/blob/main/README.md#supported-platforms)
> * there should be some reference there to React hooks support coming soon
> * support etc (eg https://github.com/ably/ably-js/blob/main/README.md#support-feedback-and-troubleshooting)

Thanks for the feedback Paddy - the changes you've mentioned are happening over at https://github.com/ably/ably-chat-js/pull/284

PR #294, opened 2024-07-16
demo-app: added presence capability

splindsay-92:
### Context

* [CHA-343]
* The demo app did not use the presence capability of the chat SDK, so this PR adds some basic usage.

### Description

* Added UserPresenceComponent and presence hook used to interactive with the presence feature
  * Component displays the list of present clientIds and their associated status
  * Component also allows the user to enter presence, update their status to `away`, and leave presence.
* Added some global and component styling.

### Checklist

* [x] QA'd by the author.

* Spin up the demo app, try pressing the buttons and check the presence list updates.

[CHA-343]: https://ably.atlassian.net/browse/CHA-343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/demo/src/hooks/usePresence.ts
+++ b/demo/src/hooks/usePresence.ts
@@ -1,0 +1,61 @@
+import * as Chat from '@ably/chat';
+import { PresenceMember } from '@ably/chat';
+import { useCallback, useEffect, useState } from 'react';
+import { useRoom } from './useRoom.ts';
+
+export const usePresence = () => {
+  const [presenceMembers, setPresenceMembers] = useState<PresenceMember[]>([]);
+  const [loading, setLoading] = useState<boolean>(true);
+  const [present, setPresent] = useState<boolean>(false);
+  const { room } = useRoom();
+
+  useEffect(() => {
+    if (!room) return;
+    room.presence.get().then((members) => {
+      setPresenceMembers(members);
+    });
+    const { unsubscribe } = room.presence.subscribe(() => {
+      room.presence.get().then((members) => {
+        setPresenceMembers(members);
+      });
+    });
+    setLoading(false);
+    return () => {
+      // cleanup after the component is unmounted
+      unsubscribe();
+    };
+  }, [room]);
+
+  const enterPresence = useCallback(
+    (data?: Chat.PresenceData) => {
+      return room.presence.enter(data).then(() => setPresent(true));
+    },
+    [room],
+  );
+
+  const leavePresence = useCallback(
+    (data?: Chat.PresenceData) => {
+      return room.presence.leave(data).then(() => {
+        setPresent(false);
+      });
+    },
+    [room],
+  );
+  const updatePresence = useCallback(
+    (data?: Chat.PresenceData) => {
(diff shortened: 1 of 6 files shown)


demo/src/hooks/usePresence.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const leavePresence = async (data?: Chat.PresenceData) => {
+    if (!present) return;
+    room.presence.leave(data).then(() => {
+      setPresent(false);
+    });
+  };
+
+  const updatePresence = async (data?: Chat.PresenceData) => {
+    await room.presence.update(data).then(() => {
```
AndyTWF, line 46: Nit-pick, no need to await here or on the others.

```ts
  const updatePresence = async (data?: Chat.PresenceData) => room.presence.update(data).then(() => setPresent(true));
```

demo/src/hooks/usePresence.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+    setLoading(false);
+    return () => {
+      // cleanup after the component is unmounted
+      unsubscribe();
+    };
+  }, [room]);
+
+  const enterPresence = (data?: Chat.PresenceData) => {
+    room.presence.enter(data).then(() => {
```
AndyTWF, line 33: This Promise should be returned so that the outside user can see its result

demo/src/hooks/usePresence.ts
splindsay-92: Good spot, I've corrected this too.

vladvelici:
I managed to quite easily reach an inconsistent state that won't resolve itself. I only ever joined with 3 browsers.

image

demo/src/hooks/usePresence.ts
splindsay-92: All the promises are now returned and handled by the outside user :)

vladvelici:
It's a react problem. Gets solved by hiding and showing again.

> Warning: Encountered two children with the same key, `john`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

We need to key the list differently.

vladvelici:
I think the `interface PresenceMember` is misleading because it's not a really a member, it's a connection. It needs to be uniquely identified by _something_ to be able to render those lists.

At the same time, I think we also want a way to get _members_ by clientId and not connection. Thoughts?

splindsay-92:
As we just discussed, I have set aside a ticket for this, [CHA-403] as it would require a breaking change to presence, which is something we need to fully discuss first :)

vladvelici:
Last thing from me: should we bold (or somhow show a `(you)`) our own username? (I'd do connectionId but we don't expose that atm)

AndyTWF (approved):
Approved pending Vlads outstanding comment

splindsay-92:
Added Vlads request :) 
Screenshot 2024-07-17 at 18 07 45

PR #297, opened 2024-07-17
template: updated PR template for react hooks.

splindsay-92:
Added a section to the definition of done for making changes to react hooks.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,6 +20,8 @@
 * [ ] Follow coding style guidelines found [here](https://github.com/ably/engineering/tree/main/best-practices).
 * [ ] TypeDoc updated (if applicable).
 * [ ] (Optional) Update documentation for new features.
+* [ ] Browser tests created (if applicable).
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -21,6 +21,12 @@
 * [ ] TypeDoc updated (if applicable).
 * [ ] (Optional) Update documentation for new features.
 
+#### For React Hooks related changes:
```
AndyTWF, line 24: We could probably collapse this into the overall checklist and just have one list for any PR on the repo - keeps things succinct and leaves it unambiguous that the first checklist still applies in the case of React.

I think the live demo app doesn't need to be mentioned explicitly - this checklist also serves as a guide for external contributors and we won't necessarily expect them to go and make changes on another repo. Arguably once hooks are released, that would be our responsibility - as it would require a release to be created here, before the package is updated outside the repo.

splindsay-92: I have made the changes :)

AndyTWF (approved):

PR #301, opened 2024-07-18
react-hooks: add useChatClient hook, related context and provider

splindsay-92:
### Context

* Provide name of JIRA ticket that this PR is related to (if applicable).
    * e.g. [CHA-101] - this will automatically link the PR to the JIRA ticket.
* Provide names of any related DRs (if applicable).
* Provide links to any related PRs

### Description

* Added `useChatClient` custom hook that allows the user to retrieve the current chat client from the nearest context.
* Added `ChatClientContext` and `ChatClientProvider`, which are necessary for using the new hook.
  * These components ensure that there is only a single instance of the chat context at any given level in the component tree. 
* Updated tsconfig to recognise the new context and provider files.

### Checklist

* [x] QA'd by the author.

* Start the demo app, check you can send reaction, messages and enter presence.

[CHA-101]: https://ably.atlassian.net/browse/CHA-101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/demo/src/App.tsx
+++ b/demo/src/App.tsx
@@ -1,5 +1,4 @@
 import { FC } from 'react';
-import { ChatClient as ChatSdk } from '@ably/chat';
 import { RoomProvider } from './containers/RoomContext';
 import { Chat } from './containers/Chat';
 import { OccupancyComponent } from './components/OccupancyComponent';
@@ -21,19 +20,14 @@ let roomId: string;
   }
 })();
 
-interface AppProps {
-  client: ChatSdk;
-}
+interface AppProps {}
 
-const App: FC<AppProps> = ({ client }) => (
-  <RoomProvider
-    client={client}
-    roomId={roomId}
-  >
-    <div style={{ display: 'flex', justifyContent: 'space-between', width: '800px' }}>
+const App: FC<AppProps> = () => (
+  <RoomProvider roomId={roomId}>
+    <div style={{ display: 'flex', justifyContent: 'space-between', width: '800px', height: '800px' }}>
       <Chat />
       <div style={{ display: 'flex', flexDirection: 'column' }}>
-        <UserPresenceComponent clientId={client.clientId} />
+        <UserPresenceComponent />
         <OccupancyComponent />
       </div>
     </div>
--- a/src/core/chat.ts
+++ b/src/core/chat.ts
@@ -42,9 +42,8 @@ export class ChatClient {
     this._clientOptions = normalizeClientOptions(clientOptions);
     const logger = makeLogger(this._clientOptions);
     this._connection = new DefaultConnection(new DefaultConnectionStatus(realtime, logger));
-
     this._rooms = new DefaultRooms(realtime, this._clientOptions, logger);
-    this._setAgent();
+    this._addAgent('chat-js');
     logger.trace(`ably chat client version ${VERSION}; initialized`);
   }
 
@@ -93,12 +92,22 @@ export class ChatClient {
   }
 
   /**
-   * Sets the agent string for the client.
+   * Adds additional agent information to the client.
+   * Used internally to add React-specific agent information.
+   * @param agent - The agent to add.
    * @internal
    */
-  private _setAgent(): void {
+  public addReactAgent(): void {
+    this._addAgent('chat-react');
+  }
+
+  /**
+   * Sets the agent string for the client.
+   * @param agent - The agent to add.
+   * @internal
+   */
+  private _addAgent(agent: string): void {
     const realtime = this._realtime as RealtimeWithOptions;
-    const agent = { 'chat-js': VERSION };
-    realtime.options.agents = { ...(realtime.options.agents ?? realtime.options.agents), ...agent };
+    realtime.options.agents = { ...(realtime.options.agents ?? realtime.options.agents), [agent]: VERSION };
   }
 }
--- a/src/react/README.md
+++ b/src/react/README.md
@@ -1,0 +1,47 @@
+# Chat React Usage Guide
+
+This document provides a brief guide on how to use custom chat components and hooks provided by the Ably Chat SDK.
+
+## ChatClientProvider
+
+This provider is used to provide the `ChatClient` instance to all child components in your React component tree.
+
+To use it, wrap your component tree with the `ChatClientProvider` and pass in the `ChatClient` instance you wish to
+use.
+
+```tsx
+import { ChatClientProvider } from '@ably/chat/react';
+import * as Ably from 'ably';
+import { LogLevel } from '@ably/chat';
+
+const realtimeClient = new Ably.Realtime({ key: 'api-key', clientId: 'clientId' });
+const chatClient = new ChatClient(realtimeClient);
+
+const App = () => {
+  return (
+    <ChatClientProvider client={chatClient}>
+      <RestOfYourApp />
+    </ChatClientProvider>
+  );
+};
+```
+
+The `ChatClient` instance will now be available to all child components in your React component tree.
+
+### useChatClient
+
+This hook allows you to access the `ChatClient` instance from your React components from the context provided by the
+`ChatClientProvider`.
+
+To use it, call the hook in your component, this will retrieve the `ChatClient` instance from the nearest
+`ChatClientProvider` in the component tree.
+
+```tsx
+import { useChatClient } from '@ably/chat/react';
+
+const MyComponent = () => {
+  const chatClient = useChatClient();
+  const clientId = chatClient.clientId;
+  return <div>Client id is: {clientId}</div>;
+};
+```
--- a/src/react/contexts/chat-client-context.tsx
+++ b/src/react/contexts/chat-client-context.tsx
@@ -1,0 +1,69 @@
+import { ChatClient } from '@ably/chat';
+import React from 'react';
+
+/**
+ * Context key to handle global context for Ably Chat Client.
+ * Only a single instance of Ably Chat context should exist to avoid issues resulted from multiple identical contexts,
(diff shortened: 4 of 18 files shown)


src/react/chat-client-context.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface GlobalThis {
+  [contextKey]?: React.Context<ChatClientContextValue>;
+}
+
+const globalObjectForContext: GlobalThis = typeof globalThis === 'undefined' ? {} : (globalThis as GlobalThis);
+
+// Ably chat context contains an object which stores all provider options indexed by provider id,
+// which is used to get options set by specific `ChatClientProvider` after calling `React.useContext`.
+export type ChatClientContextValue = Record<string, ChatClientContextProviderProps>;
```
vladvelici, line 21: I'm not sure I can follow this explanation.

Is this to be able to have more than one single ChatClient in a react app? So I could have one called `"default"` and one called `"another"`? Related to the `chatId` from use-chat-client.ts?

splindsay-92: It's the same explanation provided in the ably-js react component for providing the ably context, it would cause the same issue for us, i.e. trying to reference a room that doesn't exist in the accessed context. I think the idea is that, unless you explicitly want a different context provided, then you must understand the above issue and be happy to override it. For most users, they don't want to have to think about this, but we also don't want to remove the option for all users.

src/react/hooks/use-chat-client.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param {string} chatClientId - Identifier of the chat context.
+ * @defaultValue 'default' - This will use the default chat context.
+ *
+ * @throws {ErrorInfo} When the hook is not used within a `ChatClientProvider`.
+ *
+ * @returns {ChatClient} The chat client instance provided by the context.
+ *
+ */
+export const useChatClient = (chatClientId?: string): ChatClient => {
```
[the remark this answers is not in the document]
vladvelici: I think we need symmetry here: if we provide a way to use different chat clients by naming them in the provider, we should also provide a way to fetch them with the hook?

These are edge case uses but I can imagine one might want to
```tsx
<ChatClientProvider client={clientOne} clientId={"client_one"}>
<ChatClientProvider client={clientTwo} clientId={"client_two"}>
{ // rest of the app }
</ChatClientProvider>
</ChatClientProvider>
```

And in the rest of the app randomly use clientOne or clientTwo as they wish.

The alternative is to need to wrap components with `<ChatClientProvider>` everywhere, in which case what's the point of the `clientId` stuff anyway, because you can just get yourself a static class and shove the two clients there yourself.

It's likely I misunderstand react but I'm guessing if you do

```
<ChatClientProvider client={clientOne}>
  { // some components }
    <ChatClientProvider client={clientTwo}>
      { // here }
    </ChatClientProvider>
  { // closing some components }
</ChatClientProvider>
```
then in the `// here` you'll get `clientTwo` and not `clientOne`?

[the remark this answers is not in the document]
vladvelici: Well, after writing the above comment: what's the point of `clientId` anyway, because if the `client` is always required in the provider, `clientId` doesn't really help?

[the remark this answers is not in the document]
vladvelici: Sorry for the many comments. Last thing. If we keep it we should rename `clientId` not to confuse it with auth `clientId`.

src/react/README.md
```diff
@@ -0,0 +1,47 @@
...
+This hook allows you to access the ChatClient instance from your React components from the context provided by the
+ChatClientProvider.
+
+To use it, call the hook in your component, this will retrieve the ChatClient instance from the nearest
+ChatClientProvider in the component tree.
+
+```tsx
+import { useChatClient } from '@ably/chat/react';
+
+function SomeComponent() {
```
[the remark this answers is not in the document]
splindsay-92: Changed :)

src/react/index.ts
```diff
@@ -2,4 +2,11 @@
  * @module chat-react
  */
 
+export {
+  ChatClientContext,
```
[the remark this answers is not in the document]
splindsay-92: No, this was left in while I was trying to get typedocs to work :)

src/react/hooks/use-chat-client.ts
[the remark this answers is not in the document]
splindsay-92: Removed in both the client provider and the hook, we can add this in again should we ever want to provide support for this.

demo/src/App.tsx
```diff
@@ -21,19 +20,14 @@ let roomId: string;
...
 
-const App: FC<AppProps> = ({ client }) => (
-  <RoomProvider
-    client={client}
-    roomId={roomId}
-  >
-    <div style={{ display: 'flex', justifyContent: 'space-between', width: '800px' }}>
+const App: FC<AppProps> = () => (
+  <RoomProvider roomId={roomId}>
+    <div style={{ display: 'flex', justifyContent: 'space-between', width: '800px', height: '800px' }}>
```
vladvelici, line 27: Why the height?

src/core/chat.ts
```diff
@@ -92,13 +91,23 @@ export class ChatClient {
...
+    this._setAgent(agent);
+  }
+
   /**
    * Sets the agent string for the client.
+   * @param agent - The agent to add.
    * @internal
    */
-  private _setAgent(): void {
+  private _setAgent(agent: string): void {
```
vladvelici, line 109: nitpick: aslo rename to `_addAgent`? Seems more accurate.

src/core/config.ts
```diff
@@ -17,6 +17,17 @@ export interface ClientOptions {
...
 
+/**
+ * Configuration options for the chat client, including internal options.
+ */
+interface ClientOptionsWithInternal extends ClientOptions {
+  /**
+   * Boolean flag to indicate whether the client has been created with the React chat provider.
+   * @default false
+   */
+  useReact?: boolean;
```
vladvelici, line 28: Is this still used as it was meant to?

vladvelici (approved):
I'm approving this PR as it is. I've left minor comments in the code but I don't need to look again.

The DEFAULT_CHAT_CLIENT_ID stuff can stay or go, I'm OK with either. If it stays maybe add a comment describing why it exists/what's it good for.

demo/src/App.tsx
splindsay-92: This was to stop the screen scrolling down and leaving the app half out of view

src/core/config.ts
splindsay-92: No, now we update using an internal method, I'll remove this completely

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ably/chat",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
AndyTWF: We should bump this in the release PR rather than the feature branch

src/core/chat.ts
```diff
@@ -92,13 +91,23 @@ export class ChatClient {
     return this._clientOptions;
   }
 
+  /**
+   * Adds additional agent information to the client.
+   * Used internally to add React-specific agent information.
+   * @param agent - The agent to add.
+   * @internal
+   */
+  public addAdditionalAgent(agent: string): void {
```
AndyTWF, line 100: Perhaps as React is the only valid option here, we should have `addReactAgent`?

src/react/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```tsx
+import { ChatClientProvider } from '@ably/chat/react';
+import * as Ably from 'ably';
+import { LogLevel } from '@ably/chat';
+
+const realtimeClient = new Ably.Realtime({ key: 'api-key', clientId: 'clientId' });
+const chatClient = new ChatClient(realtimeClient);
+
+function App() {
```
AndyTWF, line 20: Shorthand arrow function

src/react/contexts/chat-client-context.tsx
```diff
@@ -0,0 +1,73 @@
+import { ChatClient } from '@ably/chat';
+import React from 'react';
+
+// Just as in the Ably React context, we need to make sure we never create more than one Ably Chat React context.
+// This might happen when exporting a context directly from a module -
+// there's a risk of creating multiple instances of the same context
+// if there are misconfigurations in module bundler or package manager on the consumer side of Ably Chat Context.
+// This can lead to problems like having a Chat Client instance added
```
AndyTWF, line 8: This comment appears to be floating - can we associate it with a line of code, or delete it?

src/react/hooks/use-chat-client.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @throws {ErrorInfo} When the hook is not used within a `ChatClientProvider`.
+ *
+ * @returns {ChatClient} The chat client instance provided by the context.
+ *
+ */
+export const useChatClient = (): ChatClient => {
+  const context = React.useContext(ChatClientContext)[DEFAULT_CHAT_CLIENT_ID];
+  if (!context) {
+    throw new Ably.ErrorInfo('hook must be used within a chat client provider', 40000, 400);
```
AndyTWF, line 20: Perhaps use `useChatClient` in the error message so it's a quick and obvious thing to find out where error is coming from

src/core/chat.ts
splindsay-92: If this is the case, I will remove the input arg since it will always be the same.

AndyTWF (approved):

PR #304, opened 2024-07-23
use ably-js EventEmitter

vladvelici:
PR in ably-js to export EventEmitter: https://github.com/ably/ably-js/pull/1819 (edit: now merged)

The reasons it uses an extra interface:
- we can define the `emit` method (it's not in ably.d.ts, not part of public API)
- as be able to use an EventMap as type argument like we did previously in Chat, it's arguably easier to work with than the stuff defined in ably.d.ts.

The reason it uses an extra class is to make the EventEmitter type easy to work with. For example, in discontinuity we export a type `type DiscontinuityEmitter = EventEmitter<DiscontinuityEventMap>;`. This doesn't work if we don't add the extra class. The alternative is to define a class here (`class DiscontinuityEmitter extends EventEmitter<DiscontinuityEventMap>`). This is the only current place where the extra class was required, but I think when we work with the EventEmitter we expect it to be a class, so this also avoids any surprises later.

This PR also removes the MockEventEmitter and uses the real thing.


src/core/utils/EventEmitter.ts
```diff
@@ -1,363 +1,15 @@
...
- * Describes the value of `this` inside the body of an {@link EventListener | `EventListener`} when invoked by an instance of {@link EventEmitter | `EventEmitter`} as the result of an event.
- *
- * @typeParam K The name of the event that this object corresponds to.
- */
-export interface EventListenerThis<K> {
-  /**
-   * The name of the event.
-   */
-  event: K;
+interface InterfaceEventEmitter<EventsMap> extends Ably.EventEmitter<Callback<EventsMap>, void, keyof EventsMap> {
```
AndyTWF, line 5: It would be good if we could typedoc everything in this file, to explain why they're needed etc.

src/core/utils/EventEmitter.ts
AndyTWF, line 1: We should rename this file to `event-emitter.ts` in line with our naming conventions

AndyTWF (commented):
Could we also remove exclude pattern from `.eslintrc.cjs` so that eslint will now run on the `utils` folder. I think in this folder we can delete the two other files, as they aren't used anywhere now.

Also the `exclude` in `tsconfig.json` would be good to get rid of :)

.eslintrc.cjs
```diff
@@ -125,7 +125,6 @@ module.exports = {
     'node_modules',
     'ably-common',
     'typedoc',
-    'src/core/utils',
     'test/core/utils',
```
AndyTWF, line 128: Can remove this one too

tsconfig.json
```diff
@@ -7,7 +7,6 @@
     "./test/**/*.d.ts",
   ],
   "exclude": [
-    "src/core/utils/**/*",
     "test/core/utils/**/*",
   ],
```
AndyTWF, line 11: Might as well take this out too

AndyTWF (approved):
:shipit:

PR #306, opened 2024-07-25
demo app: simplify css

vladvelici:


AndyTWF (commented):
Now that we have the connection status box from #310 - could we make it so that said box is always shown just transparent? Would mean our messages don't jump around in response to connection state changing

vladvelici:
@AndyTWF I've made the connection component stick on top so it won't make content jump when it's shown.

AndyTWF (approved):

PR #318, opened 2024-08-06
Async room initialisation

vladvelici:
### Context

The problem is also described at [CHADR-049] and (loosely) tracked at [CHA-382](https://ably.atlassian.net/browse/CHA-382).

### Description

Our current public API allows users to get a room that is currently being released, which is undesirable:

```typescript
// options emitted for brevity
const r1 = rooms.get("abc");
rooms.release("abc").then(() => { console.log("released abc"); });
const r2 = rooms.get("abc"); // this is the room being released, unusable right now but we want it to work
// -> released abc
const r3 = rooms.get("abc") // this is now usable
```

The changes in this PR make `r2` above to be a usable room, which will have status `initializing` until the previous room (from `r1`) finishes releasing. No realtime channels are created in `initializing` state. Realtime channels are created when the previous room finishes releasing, and after channel creating completes (for all configured features) the room becomes `initialized` (which was the previous default status).

#### Summary of changes

1. Introduce new RoomLifecycle status `initializing` which is now the default status when a new room is created.
2. Make all room features receive a promise that they must await before creating any realtime channels.
3. Each feature's `channel` getter is now for a promise of a realtime channel. The room only changes status to `Initialized` and starts the usual lifecycle when all channel promises are resolved.
4. `RoomLifecycle` enum is now exported as value, not as type.
5. If a room is released before it started to initialize, then it will never initialize. Methods like `room.attach()` return a rejected promise explaining the room was released before initialization.

### Checklist

* [ ] QA'd by the author.

- Run the tests
- Read the code around room.ts and channel initialisation in features carefully
- Read added comments in tests where I had to use tricks like waiting a small amount of time

[CHA-382]: https://ably.atlassian.net/browse/CHA-382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/discontinuity.ts
+++ b/src/core/discontinuity.ts
@@ -7,6 +7,11 @@ import EventEmitter from './utils/event-emitter.js';
  */
 export interface HandlesDiscontinuity {
   /**
+   * The channel that this object is associated with.
+   */
+  get channel(): Promise<Ably.RealtimeChannel>;
+
+  /**
    * Called when a discontinuity is detected on the channel.
    * @param reason The error that caused the discontinuity.
    */
--- a/src/core/discontinuity.ts
+++ b/src/core/discontinuity.ts
@@ -7,11 +7,6 @@ import EventEmitter from './utils/event-emitter.js';
  */
 export interface HandlesDiscontinuity {
   /**
-   * The channel that this object is associated with.
-   */
-  get channel(): Ably.RealtimeChannel;
-
-  /**
    * Called when a discontinuity is detected on the channel.
    * @param reason The error that caused the discontinuity.
    */
--- a/src/core/discontinuity.ts
+++ b/src/core/discontinuity.ts
@@ -7,7 +7,8 @@ import EventEmitter from './utils/event-emitter.js';
  */
 export interface HandlesDiscontinuity {
   /**
-   * The channel that this object is associated with.
+   * A promise of the channel that this object is associated with. The promise
+   * is resolved when the feature has finished initializing.
    */
   get channel(): Promise<Ably.RealtimeChannel>;
 
--- a/src/core/messages.ts
+++ b/src/core/messages.ts
@@ -243,7 +243,9 @@ export class DefaultMessages
     this._channel = initAfter.then(() => this._makeChannel(roomId, realtime));
 
     // Catch this so it won't send unhandledrejection global event
-    this._channel.catch((error : unknown) => { logger.debug('Messages: channel initialization canceled', { roomId, error }); });
+    this._channel.catch((error: unknown) => {
+      logger.debug('Messages: channel initialization canceled', { roomId, error });
+    });
 
     this._chatApi = chatApi;
     this._clientId = clientId;
--- a/src/core/messages.ts
+++ b/src/core/messages.ts
@@ -197,7 +197,7 @@ export interface Messages extends EmitsDiscontinuities {
    *
    * @returns the realtime channel
    */
-  get channel(): Ably.RealtimeChannel;
+  get channelPromise(): Promise<Ably.RealtimeChannel>;
 }
 
 /**
@@ -208,7 +208,7 @@ export class DefaultMessages
   implements Messages, HandlesDiscontinuity, ContributesToRoomLifecycle
 {
   private readonly _roomId: string;
-  private readonly _channel: Ably.RealtimeChannel;
+  private readonly _channelPromise: Promise<Ably.RealtimeChannel>;
   private readonly _chatApi: ChatApi;
   private readonly _clientId: string;
   private readonly _listenerSubscriptionPoints: Map<
@@ -228,32 +228,48 @@ export class DefaultMessages
    * @param clientId The client ID of the user.
    * @param logger An instance of the Logger.
    */
-  constructor(roomId: string, realtime: Ably.Realtime, chatApi: ChatApi, clientId: string, logger: Logger) {
+  constructor(
+    roomId: string,
+    realtime: Ably.Realtime,
+    chatApi: ChatApi,
+    clientId: string,
+    logger: Logger,
+    initAfter: Promise<void>,
+  ) {
     super();
     this._roomId = roomId;
-    this._channel = getChannel(messagesChannelName(roomId), realtime);
-    addListenerToChannelWithoutAttach({
-      listener: this._processEvent.bind(this),
-      events: [MessageEvents.Created],
-      channel: this._channel,
+
+    this._channelPromise = initAfter.then(() => {
+      const channel = getChannel(messagesChannelName(roomId), realtime);
+
+      addListenerToChannelWithoutAttach({
+        listener: this._processEvent.bind(this),
+        events: [MessageEvents.Created],
+        channel: channel,
+      });
+
+      // Handles the case where channel attaches and resume state is false. This can happen when the channel is first attached,
+      // or when the channel is reattached after a detach. In both cases, we reset the subscription points for all listeners.
+      channel.on('attached', (message) => {
+        this._handleAttach(message.resumed);
+      });
+      // Handles the case where an update message is received from a channel after a detach and reattach.
+      channel.on('update', (message) => {
+        if (message.current === 'attached' && message.previous === 'attached') {
+          this._handleAttach(message.resumed);
+        }
+      });
+
+      return channel;
     });
+
+    // catch this so it won't send unhandledrejection global event
+    this._channelPromise.catch(() => {});
 
     this._chatApi = chatApi;
     this._clientId = clientId;
     this._logger = logger;
     this._listenerSubscriptionPoints = new Map<MessageListener, Promise<{ fromSerial: string }>>();
-
-    // Handles the case where channel attaches and resume state is false. This can happen when the channel is first attached,
-    // or when the channel is reattached after a detach. In both cases, we reset the subscription points for all listeners.
-    this._channel.on('attached', (message) => {
-      this._handleAttach(message.resumed);
-    });
-    // Handles the case where an update message is received from a channel after a detach and reattach.
-    this._channel.on('update', (message) => {
-      if (message.current === 'attached' && message.previous === 'attached') {
-        this._handleAttach(message.resumed);
-      }
-    });
   }
 
   /**
@@ -328,10 +344,10 @@ export class DefaultMessages
   private async _resolveSubscriptionStart(): Promise<{
     fromSerial: string;
   }> {
-    const channelWithProperties = this._getChannelProperties();
+    const channelWithProperties = await this._getChannelProperties();
 
     // If we are attached, we can resolve with the channelSerial
-    if (this._channel.state === 'attached') {
+    if (channelWithProperties.state === 'attached') {
       if (channelWithProperties.properties.channelSerial) {
         return { fromSerial: channelWithProperties.properties.channelSerial };
       }
@@ -342,11 +358,15 @@ export class DefaultMessages
     return this._subscribeAtChannelAttach();
   }
 
-  private _getChannelProperties(): Ably.RealtimeChannel & {
-    properties: { attachSerial: string | undefined; channelSerial: string | undefined };
-  } {
+  private async _getChannelProperties(): Promise<
+    Ably.RealtimeChannel & {
+      properties: { attachSerial: string | undefined; channelSerial: string | undefined };
+    }
+  > {
     // Get the attachSerial from the channel properties
-    return this._channel as Ably.RealtimeChannel & {
+    // const channel = await this._channelPromise;
+    const channel = await this._channelPromise;
+    return channel as Ably.RealtimeChannel & {
       properties: {
         attachSerial: string | undefined;
         channelSerial: string | undefined;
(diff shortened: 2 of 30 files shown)


vladvelici:
Found a potential thing that might need to be addressed in this situation:

1. a new room gets created
2. p = room.attach() called
3. previous room finishes initializing
4. features do not finish creating channels
5. room.release() gets called
6. features finish initializing
7. now we want to just release without attaching anymore (just reject the attachment promise), but I think what will happen is an attach/detach cycle

src/core/occupancy.ts
```diff
@@ -110,15 +110,23 @@ export class DefaultOccupancy
...
+      addListenerToChannelWithoutAttach({
+        listener: this._internalOccupancyListener.bind(this),
+        events: ['[meta]occupancy'],
+        channel: channel,
+      });
+      return channel;
     });
+
+    // catch this so it won't send unhandledrejection global event
+    this._channel.catch(() => void 0);
```
splindsay-92, line 128: What events lead to the error being thrown and not caught by using `initAfter.then().catch()` ?

src/core/occupancy.ts
```diff
@@ -110,15 +110,23 @@ export class DefaultOccupancy
    * @param chatApi An instance of the ChatApi.
    * @param logger An instance of the Logger.
    */
```
splindsay-92, line 112: Missing `@param initAfter`

src/core/presence.ts
```diff
@@ -204,7 +204,14 @@ export class DefaultPresence
    * A channel can have multiple connections using the same clientId.
    * @param logger An instance of the Logger.
```
splindsay-92, line 205: Missing param here also :)

src/core/room.ts
```diff
@@ -118,8 +119,16 @@ export class DefaultRoom implements Room {
    * @param realtime An instance of the Ably Realtime client.
    * @param chatApi An instance of the ChatApi.
    * @param logger An instance of the Logger.
+   * @param initAfter The room will wait for this promise to finish before initializing
    */
-  constructor(roomId: string, options: RoomOptions, realtime: Ably.Realtime, chatApi: ChatApi, logger: Logger) {
+  constructor(
```
splindsay-92, line 124: This is quite a bit of code in the constructor, is it worth pulling some of it out into functions to reduce the cognitive complexity a bit? :)

src/core/room.ts
```diff
@@ -107,8 +107,9 @@ export class DefaultRoom implements Room {
   private readonly _occupancy?: DefaultOccupancy;
   private readonly _logger: Logger;
   private readonly _status: DefaultStatus;
-  private readonly _lifecycleManager: RoomLifecycleManager;
-  private readonly _finalizer: () => Promise<void>;
+  private _lifecycleManager?: RoomLifecycleManager;
+  private _finalizer: () => Promise<void>;
+  private _asyncOpsAfter: Promise<void>;
```
splindsay-92, line 112: This could be readonly I think

src/core/room.ts
```diff
@@ -129,53 +138,176 @@ export class DefaultRoom implements Room {
...
     this._status = new DefaultStatus(logger);
 
+    // This function gets called if release() is called before initialization
+    // starts. It allows for the room to not be initialized at all since it
+    // won't be needed.
+    let stopInitializingFeatures: (() => void) | undefined;
+
+    // This promise is the same as initAfter but it gets rejected if release()
+    // is called before initialization starts. This make sure that room
+    // features will not permanently hang waiting for this promise to resolve.
```
splindsay-92, line 148:
```suggestion
    // This promise is the same as initAfter, but it gets rejected if release()
    // is called before initialization starts. This makes sure that room
    // features will not permanently hang waiting for this promise to resolve.
```

splindsay-92 (commented):
Looks good so far, I'm still looking through it, but wanted to get a few comments up :)

src/core/messages.ts
```diff
@@ -228,32 +228,48 @@ export class DefaultMessages
...
+      channel.on('update', (message) => {
+        if (message.current === 'attached' && message.previous === 'attached') {
+          this._handleAttach(message.resumed);
+        }
+      });
+
+      return channel;
     });
 
+    // catch this so it won't send unhandledrejection global event
```
splindsay-92, line 266: little nitpick with this and a couple other comments to keep case consistency :) 
```suggestion
    // Catch this so it won't send unhandledrejection global event
```

src/core/messages.ts
```diff
@@ -228,32 +228,48 @@ export class DefaultMessages
...
+  ) {
     super();
     this._roomId = roomId;
-    this._channel = getChannel(messagesChannelName(roomId), realtime);
-    addListenerToChannelWithoutAttach({
-      listener: this._processEvent.bind(this),
-      events: [MessageEvents.Created],
-      channel: this._channel,
+
+    this._channel = initAfter.then(() => {
```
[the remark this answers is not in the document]
vladvelici: Good idea

src/core/messages.ts
```diff
@@ -228,32 +228,48 @@ export class DefaultMessages
...
+        if (message.current === 'attached' && message.previous === 'attached') {
+          this._handleAttach(message.resumed);
+        }
+      });
+
+      return channel;
     });
 
+    // catch this so it won't send unhandledrejection global event
+    this._channel.catch(() => void 0);
```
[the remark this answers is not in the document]
vladvelici: The only reason we catch this here is so that if it's not used anywhere where it gets caught, there's a global error happening [[1](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event)]. I didn't know about this until writing this code.

If a release happens before this resolves, then this promise gets rejected. As a result, all ops waiting for this promise will be rejected, which makes sense to me.

A log is a good idea, I'll go for a debug.

src/core/occupancy.ts
vladvelici: if we do
```
this._channel = initAfter.then(...).catch(...)
```
then this._channel will be a promise that always _resolves_ (in case of catch it'll resolve to whatever that returns, likely `undefined`), and is never rejected (assuming the catch block doesn't return `Promise.reject(...)`). We don't want that. We want it to be rejected if release happened before init.

So to _propagate_ the rejection we just don't catch it. But that means the global unhandled promise thing gets triggered if this._channel didn't have a `catch` registered anywhere else in code at the time of rejection.

src/core/room.ts
```diff
@@ -129,53 +138,176 @@ export class DefaultRoom implements Room {
...
-    this._finalizer = async () => {
-      // Cycle the channels in the feature and release them from the realtime client
-      if (finalized) {
-        this._logger.debug('Room.finalizer(); already finalized');
-        return;
+    // Wait for features to finish initializing and then finish initializing
+    // the room. Set _asyncOpsAfter to the promise that waits for the room to
+    // finish initializing. This promise is awaited before performing any async
+    // operations at room level (attach and detach).
+    this._asyncOpsAfter = initFeaturesAfter.then(() => {
```
splindsay-92, line 241: I think this should be moved out of the constructor too please, there's quite a lot going on here :)

src/core/room.ts
```diff
@@ -129,53 +138,176 @@ export class DefaultRoom implements Room {
...
+        setFinalizerFunc = resolve;
+      });
+      this._finalizer = () => {
+        return finalizerFuncPromise.then((f) => {
+          return f();
+        });
+      };
+
+      // Setup all contributors with resolved channels
+      interface ContributorWithChannel {
```
splindsay-92, line 259: Another nitpick, should we move this interface outside the constructor and document it? Personally I think its better to define interfaces at the top level, or do you think it's best here?

src/core/room.ts
```diff
@@ -257,17 +389,17 @@ export class DefaultRoom implements Room {
   /**
    * @inheritdoc Room
    */
-  async attach(): Promise<void> {
+  async attach() {
     this._logger.trace('Room.attach();');
-    return this._lifecycleManager.attach();
+    return this._asyncOpsAfter.then(() => this._lifecycleManager?.attach());
```
splindsay-92, line 394: Do we want to ensure the `_lifecycleManager` is not undefined here? We could then return a better defined error, what do you think?

src/core/room-lifecycle-manager.ts
```diff
@@ -11,9 +11,10 @@ import { DefaultStatus, NewRoomStatus, RoomLifecycle, RoomStatusChange } from '.
  */
 export interface ContributesToRoomLifecycle extends HandlesDiscontinuity {
   /**
-   * Gets the channel on which the feature operates.
+   * Gets the channel on which the feature operates. This promise is never
+   * rejected except in the case where room initialization is cancelled.
```
splindsay-92, line 15: small typo - `canceled`

src/core/room-lifecycle-manager.ts
```diff
@@ -28,6 +29,11 @@ export interface ContributesToRoomLifecycle extends HandlesDiscontinuity {
   get detachmentErrorCode(): ErrorCodes;
 }
 
+export interface RoomLifecycleContributor {
```
splindsay-92, line 32: This needs some type docs :)

test/core/room.test.ts
```diff
@@ -134,7 +165,215 @@
...
+    expect(context.realtime.channels.get).not.toHaveBeenCalled();
+
+    resolve();
+
+    // allow a tick to happen
+    await new Promise((res) => setTimeout(res, 0));
+
+    const events : string[] = [];
+
+    // must sill be initializing since messages channel is not yet initialized
```
splindsay-92, line 353: Small typo here

test/core/room.test.ts
```diff
@@ -134,7 +165,215 @@ describe('Room', () => {
...
+    expect(room.status.current).toBe(RoomLifecycle.Initializing);
+    const releasePromise = (room as DefaultRoom).release();
+  
+    // expect the release promise to be different than initAfter now, because
```
splindsay-92, line 357:
```suggestion
    // expect the release promise to be different to initAfter now, because
```

splindsay-92:
Great work so far! I'll wait till the tests/changes are all done and give it another look over :)

src/core/discontinuity.ts
```diff
@@ -6,11 +6,6 @@ import EventEmitter from './utils/event-emitter.js';
  * Represents an object that has a channel and therefore may care about discontinuities.
```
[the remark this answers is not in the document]
vladvelici: I'm adding back the channel but as a Promise. Seems to make more sense when reading it.

src/core/room-lifecycle-manager.ts
vladvelici: Done, and added cspell config to catch `"cancelled"`

src/core/rooms.ts
```diff
@@ -103,11 +107,50 @@ export class DefaultRooms implements Rooms {
...
+      // this callback is at least as high as the current count.
+      //
+      // This is to handle the case where multiple release calls
+      // are made in quick succession. We only want to remove the
+      // room from the releasing map if the last ongoing release
+      // finished.
+      //
+      // The release callbacks are actually happening in the order
+      // in which release() methods are called due to passing
+      // waitForMe in rooms.get().
```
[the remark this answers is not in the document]
vladvelici: I'll update the comment.

src/core/room.ts
vladvelici: removed it entirely

src/core/room.ts
vladvelici: I've refactored some bits. It still looks hard to read but moved lots of logic outside the constructor.

src/core/room.ts
splindsay-92: Ah I see, I suppose it's not necessary then if `_lifecycleManager` is guaranteed to be there after `_asyncOpsAfter` resolves :)

src/core/rooms.ts
```diff
@@ -103,11 +107,46 @@ export class DefaultRooms implements Rooms {
...
+    const releasing = this._releasing.get(roomId);
+
+    // if we don't have the room
+    if (!room) {
+      // ... and it's currently releasing wait for the same promise
+      if (releasing) {
+        return releasing.promise;
+      }
+
+      // if not releasing, nothing else to do
```
splindsay-92, line 119: Small nitpicks again,
`// if the room is not releasing, there is nothing else to do`

src/core/typing.ts
```diff
@@ -71,7 +71,7 @@ export interface Typing extends EmitsDiscontinuities {
    * Get the name of the realtime channel underpinning typing events.
    * @returns The name of the realtime channel.
```
splindsay-92, line 72: Same here with updating to return a Promise

splindsay-92 (commented):
Looking really good, only a couple comments, mostly just me nitpicking!

src/core/occupancy.ts
```diff
@@ -48,7 +48,7 @@ export interface Occupancy extends EmitsDiscontinuities {
    *
    * @returns The underlying Ably channel for occupancy events.
```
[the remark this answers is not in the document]
vladvelici: I've updated the channel comments in all feature interfaces to say it's a promise.

splindsay-92 (approved):
Just one last tiny comment from me and then I think it's great! I'll approve as to not block you once the change is made since it's only a small wording change.

PR #321, opened 2024-08-13
react-hooks: Add useTyping hook.

splindsay-92:
### Context

* [CHA-386]
* As part of the react hooks development, a new hook to interact with the `Typing` instance of a room was required.

### Description

- Added the `useTyping` hook to provide access to the `Typing` instance and associated functionality.
  -  Start and stop events can be triggered by calling the methods returned from the hook.
  - A listener can also be provided to receive any incoming events that the `Typing` instance of the room receives.
  - The hook returns a state, that it keeps up-to-date, containing the set of current typers.
 
- Updated the demo app to use the new hook, removed old functionality.
- Added a section on this hooks usage to the README.md.

### Checklist

* [x] QA'd by the author.

1. Spin up a local instance of the demo app.
2. Open up two tabs and start typing on one.
3. Check a typing indicators appears and then disappears after you stop typing. (Should be a short delay) 

[CHA-386]: https://ably.atlassian.net/browse/CHA-386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/react/hooks/use-typing.ts
```diff
@@ -0,0 +1,90 @@
...
+  const stop = useCallback(() => room.typing.stop(), [room.typing]);
+
+  return {
+    connectionStatus,
+    connectionError,
+    roomStatus,
+    roomError,
+    start,
+    stop,
+    currentlyTyping,
```
AndyTWF, line 87: Should `currentlyTyping` perhaps be a property, that automatically gets updated by some internal listener, similar to what we have with occupancy?

AndyTWF (commented):
Can we add a README entry here?

src/react/hooks/use-typing.ts
splindsay-92: It can indeed, I'll add the same throttle capability as the `useOccupancy` hook :)

test/react/hooks/use-typing.test.tsx
```diff
@@ -0,0 +1,212 @@
+import { ConnectionLifecycle, Room, RoomLifecycle, TypingListener } from '@ably/chat';
+import { act, renderHook } from '@testing-library/react';
```
AndyTWF, line 2: Can we get an integration test that runs end-end using multiple clients and a sandbox?

demo/src/containers/Chat/Chat.tsx
```diff
@@ -2,16 +2,18 @@ import { useCallback, useEffect, useRef } from 'react';
...
-import { useChatConnection } from '@ably/chat/react';
+import { useChatConnection, useTyping } from '@ably/chat/react';
 import { ConnectionLifecycle } from '@ably/chat';
 
 export const Chat = () => {
   const { loading, clientId, messages, sendMessage } = useMessages();
-  const { startTyping, stopTyping, typers } = useTypingIndicators();
+
+  const { start, stop, currentlyTyping } = useTyping({
+    updateInterval: 1000,
```
vladvelici, line 15: If we keep the throttling, maybe we should think about the name of this parameter. `maxUpdateInterval`? Also, not a big deal but there's a potential confusion with the `typingTimeout`, the timer that calls `stop()` for you.

demo/src/containers/Chat/Chat.tsx
```diff
@@ -27,6 +29,14 @@ export const Chat = () => {
     [sendMessage],
   );
 
+  const handleStartTyping = useCallback(() => {
```
vladvelici, line 32: These are already callbacks in the hook, right? So they shouldn't need double-wrapping?

vladvelici (commented):
Left very minor comments.

vladvelici (approved):
I only now saw that we split this into two hooks. This one's good to go!

splindsay-92:
> I only now saw that we split this into two hooks. This one's good to go!

@vladvelici I think you mean Presence right?

vladvelici:
> > I only now saw that we split this into two hooks. This one's good to go!
> 
> @vladvelici I think you mean Presence right?

Yes, I did mean Presence. Sorry!

src/react/hooks/use-typing.ts
```diff
@@ -0,0 +1,100 @@
...
+    onStatusChange: params?.onConnectionStatusChange,
+  });
+  const { room, roomError, roomStatus } = useRoom({
+    onStatusChange: params?.onRoomStatusChange,
+  });
+
+  const [currentlyTyping, setCurrentlyTyping] = useState<Set<string>>(new Set());
+
+  useEffect(() => {
+    const { unsubscribe } = room.typing.subscribe((event) => {
```
vladvelici, line 60: Should this call `get()` before it subscribes to make sure the set is up-to-date on page/room load even if there are no typing events coming?

splindsay-92: Yeh that's a good point, I'll add this as part of the mount process and do the same for presence :)

src/react/hooks/use-typing.ts
```diff
@@ -0,0 +1,100 @@
...
+      setCurrentlyTyping(event.currentlyTyping);
+    });
+    return () => {
+      unsubscribe();
+    };
+  }, [room]);
+
+  // if provided, subscribes the user provided onDiscontinuity listener
```
vladvelici, line 68:
```suggestion
  // if provided, subscribes the user-provided onDiscontinuity listener
```

AndyTWF (approved):

PR #322, opened 2024-08-14
react-hooks: Add usePresence hook

splindsay-92:
### Context

* [CHA-387]
* As part of the react hooks development, a new hook to interact with the `Presence` instance of a room was required.

### Description

- Added the `usePresence` hook to the Chat React library. This allows the user to interact with the underlying `Presence` instance, as well as easily update and leave presence.
- A user is entered into presence on rendering the hook.
- Updated the demo app to use the new hook, this includes updating and leaving presence.
  - Entering presence is now automatic on calling the hook.
- Updated the readme to document hook usage.
- Added unit tests for the new hook.

### Checklist

* [x] QA'd by the author.

1. Start a local demo app
2. On loading the app, you should see your clientId added to the presence list.
3. Check opening multiple tabs adds the users to the presence list.
4. Check unmounting the component leaves presence and the presence lists of other clients are updated.
5. Check the leave button leaves presence.
6. Check the update button sets the presence data.

[CHA-387]: https://ably.atlassian.net/browse/CHA-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core/presence.ts
```diff
@@ -123,35 +123,35 @@ export interface Presence extends EmitsDiscontinuities {
...
-   * @returns {Promise<PresenceMessage[]>} or upon failure, the promise will be rejected with an [[Ably.ErrorInfo]] object which explains the error.
+   * @returns {Promise<PresenceMessage[]>} or upon failure, the promise will be rejected with an {@link Ably.ErrorInfo} object which explains the error.
    */
   get(params?: Ably.RealtimePresenceParams): Promise<PresenceMember[]>;
 
   /**
    * Method to check if user with supplied clientId is online
    * @param {string} clientId - The client ID to check if it is present in the room.
-   * @returns {Promise<{boolean}>} or upon failure, the promise will be rejected with an {@link ErrorInfo} object which explains the error.
+   * @returns {Promise<{boolean}>} or upon failure, the promise will be rejected with an {@link Ably.ErrorInfo} object which explains the error.
```
AndyTWF, line 133: Does this break the automatic linking in typedoc? I vaguely remember having to export `ErrorInfo` in `index.ts` because otherwise it just linked to nothing

src/react/hooks/use-presence.ts
```diff
@@ -0,0 +1,128 @@
...
+    // wait until the room is attached before attempting to enter
+    if (roomStatus !== RoomLifecycle.Attached) return;
+    void room.presence.enter(dataRef.current).then(() => {
+      setIsPresent(true);
+    })
+
+    return () => {
+      // ensure we are still in an attached state before attempting to leave;
+      // a presence.leave call will produce an exception otherwise.
+      if (roomStatusRef.current === RoomLifecycle.Attached) {
```
AndyTWF, line 85: We also need to check the connection state here, as an inactive connection will cause an exception:

https://github.com/ably/ably-js/blob/8ba8ec29f521b18ffd16bc4fcb33004c64a7d706/src/platform/react-hooks/src/hooks/usePresence.ts#L78

src/core/presence.ts
splindsay-92: Yeh ignore this, my IDE threw up errors and I meant to check it once the docs worked (currently broken by the rooms work), but I just checked main and it was just my IDE playing tricks.

src/react/hooks/use-presence.ts
```diff
@@ -0,0 +1,160 @@
...
+  // enter the room when the hook is mounted
+  useEffect(() => {
+    const canJoinPresence = roomStatus === RoomLifecycle.Attached && !INACTIVE_CONNECTION_STATES.has(connectionStatus);
+    const canLeavePresence =
+      roomStatusAndConnectionStatusRef.current.roomStatus === RoomLifecycle.Attached &&
+      !INACTIVE_CONNECTION_STATES.has(roomStatusAndConnectionStatusRef.current.connectionStatus);
+
+    // wait until the room is attached before attempting to enter, and ensure the connection is active
+    if (!canJoinPresence) return;
+    room.presence
```
vladvelici, line 95: I've seen some discussion before -- ignore if already resolved. We enter presence by default when this element is in scope? We can't just use it to list the users who are in presence without joining?

I think there's some room for use cases where you're not allowed to join presence (not VIP?) but you can see who's there (the list of VIPs)?

vladvelici (commented):
Looks good. Left one question.

src/react/hooks/use-presence.ts
splindsay-92: That's correct, @AndyTWF and I discussed this and he thought the expectation here by users would be the hook automatically joins presence.

src/react/README.md
```diff
@@ -67,3 +67,40 @@ const MyComponent = () => {
   return <div>Connection status is: {currentStatus}</div>;
 };
 ```
+
+## usePresence
+
+This hook allows you to access the `Presence` instance of a specific room from the nearest `ChatRoomProvider` in the
```
AndyTWF, line 73: I think we should be more specific earlier on in the documentation that this hook relates to enter/update operations with presence. I have to go down 3 paragraphs to deduce this - which I think is probably too far.

We should also direct people to the appropriate hook for listening to presence events.

src/react/hooks/use-presence.ts
```diff
@@ -0,0 +1,160 @@
...
+  }, [params?.data]);
+
+  useEffect(() => {
+    // Update the ref when roomStatus changes
+    roomStatusAndConnectionStatusRef.current = { roomStatus, connectionStatus };
+  }, [roomStatus, connectionStatus]);
+
+  // enter the room when the hook is mounted
+  useEffect(() => {
+    const canJoinPresence = roomStatus === RoomLifecycle.Attached && !INACTIVE_CONNECTION_STATES.has(connectionStatus);
```
AndyTWF, line 88: We need a check here to see if a user has explicitly left presence - the `useEffect` will run every time the room status changes, but if they've explicitly left presence, we shouldn't automatically re-enter them.

Either that, or we remove the `leave` method and handle the lifecycle of being present by the lifecycle of the hook (i.e. you can update your presence data whenever you like, but if you don't want to remain present, you should stop calling the hook).

typedoc.json
```diff
@@ -3,7 +3,7 @@
   "entryPoints": ["./src/core/index.ts", "./src/react/index.ts"],
   "out": "typedoc/generated",
   "readme": "typedoc/landing-page.md",
-  "treatWarningsAsErrors": true,
+  "treatWarningsAsErrors": false,
```
AndyTWF, line 6: Why have we changed this? Treating warnings as errors helps us identify things that are missing documentation.

src/react/hooks/use-presence.ts
splindsay-92: I think the idea of removing it here all together keeps us more in line with how ably-js do things, so I will go with that approach. Also, it feel a bit more intuitive that way too.

typedoc.json
splindsay-92: Ahh yes, this was done during testing while none of the typedocs worked, I missed setting this one back, I'll correct that now.

src/react/hooks/use-presence.ts
```diff
@@ -0,0 +1,151 @@
...
+    [room],
+  );
+
+  return {
+    connectionStatus,
+    connectionError,
+    roomStatus,
+    roomError,
+    update,
+    isPresent,
```
AndyTWF, line 147: With removal of `leave`, does `isPresent` make much sense anymore, as it will only be driven by normal lifecycle events?

AndyTWF (approved):

PR #325, opened 2024-08-16
async room initialization with sync rooms.get

vladvelici:
This is an early draft for the async room initialization. I've added a new status, `RoomLifecycle.Initializing`, to show that the room didn't start init yet.

Please have a look at:
- `room.ts` (start here)
- `rooms.ts`
- `room-lifecycle-manager.ts`
- and one of the features like `messages.ts`, `typing-indicators.ts` etc (they're all refactored the same way)

### DIFF
--- a/src/core/messages.ts
+++ b/src/core/messages.ts
@@ -197,7 +197,7 @@ export interface Messages extends EmitsDiscontinuities {
    *
    * @returns the realtime channel
    */
-  get channel(): Ably.RealtimeChannel;
+  get channelPromise(): Promise<Ably.RealtimeChannel>;
 }
 
 /**
@@ -208,7 +208,7 @@ export class DefaultMessages
   implements Messages, HandlesDiscontinuity, ContributesToRoomLifecycle
 {
   private readonly _roomId: string;
-  private readonly _channel: Ably.RealtimeChannel;
+  private readonly _channelPromise: Promise<Ably.RealtimeChannel>;
   private readonly _chatApi: ChatApi;
   private readonly _clientId: string;
   private readonly _listenerSubscriptionPoints: Map<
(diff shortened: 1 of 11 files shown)


src/core/messages.ts
```diff
@@ -396,8 +411,8 @@ export class DefaultMessages
   /**
    * @inheritdoc Messages
    */
-  get channel(): Ably.RealtimeChannel {
```
AndyTWF, line 399: I think you could still call this `channel` - its not usually common to add `promise` to the method name

AndyTWF: Ditto for the others

src/core/room.ts
```diff
@@ -129,53 +138,146 @@ export class DefaultRoom implements Room {
...
+    // won't be needed.
+    let stopInitializingFeatures: (() => void) | undefined;
+
+    // This promise is the same as initAfter but it gets rejected if release()
+    // is called before initialization starts. This make sure that room
+    // features will not permanently hang waiting for this promise to resolve.
+    //
+    // This promise is passed down to all features to wait before starting to
+    // create or use any realtime channels.
+    const initFeaturesAfter = new Promise<void>((resolve, reject) => {
```
AndyTWF, line 152: A note here - any async ops as part of feature initialisation will need to be `await` or become synchronous, we need them all to finish before this promise resolves

vladvelici: Good point! All feature init is sync now and runs after this promise resolves. And because the channels are now promises they could be async and those are awaited below (lines 240, 252)

vladvelici:
Closing this PR. The work was followed up in https://github.com/ably/ably-chat-js/pull/318.

PR #333, opened 2024-09-02
CDN Bundle Fixes

AndyTWF:
### Context

We were having some issues with using ably-js and ably-chat-js from the CDN, as the former doesn't export an ESM module (and the later tries to import it).

### Description

- Update the Vite library names for Chat and React to be `AblyChat` and `AblyChatReact` respectively. This means that users can do `const chatClient = new AblyChat.ChatClient()` instead of `const chatClient = new window['ably-chat-js'].ChatClient()` in the browser.
- Update the README instructions for using the CDN to point to the UMD variant of Chat, which does work with ably-js.
- Update the React dependencies to make `react-dom` and `react/jsx-runtime` externals - requiring the consumer to provide them and avoiding conflicts.

### Checklist

* [x] QA'd by the author.

Minimal working example:

```html
<script src="https://cdn.ably.com/lib/ably.min-2.js"></script>
<script src="./path-to-ably-chat-js-umd-variant"></script>
<script>
  const realtime = new Ably.Realtime({
          key: 'YOUR KEY HERE',
          clientId: 'client_' + client_id,
        });
  const chatClient = new AblyChat.ChatClient(realtime);
</script>
```

If the above doesn't error, you're good!

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,13 @@ npm install @ably/chat
 For browsers, you can also include the Chat SDK directly into your HTML:
 
 ```html
-<script src="https://cdn.ably.com/lib/ably-chat-0.js"></script>
+<!-- Ably Chat also requires the core Ably SDK to be available -->
+<script src="https://cdn.ably.com/lib/ably.min-2.js"></script>
+<script src="https://cdn.ably.com/lib/ably-chat.umd.cjs-0.js"></script>
+<script>
+  const realtime = new Ably.Realtime({key: 'your-ably-key'})
+  const chatClient = new AblyChat.ChatClient(realtime);
+</script>
 ```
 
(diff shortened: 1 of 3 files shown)


src/react/README.md
```diff
@@ -2,6 +2,31 @@
...
+```html
+<!-- Ably Chat also requires the core Ably SDK to be available -->
+<script src="https://cdn.ably.com/lib/ably.min-2.js"></script>
+<!-- You also need the base SDK -->
+<script src="https://cdn.ably.com/lib/ably-chat.umd.cjs-0.js"></script>
+<!-- Include React base library -->
+<script src="https://unpkg.com/react@18/umd/react.production.min.js"></script>
+<script src="https://cdn.ably.com/lib/ably-chat-react.umd.cjs-0.js"></script>
+<script>
+  const chatClient = new AblyChat.ChatClient(realtime);
```
splindsay-92, line 24: Should we include some example calling `AblyChatReact` here?

splindsay-92 (approved):
Tested locally and it works, nicely done!

src/react/README.md
AndyTWF: In hindsight... I'm going to get rid of React on the CDN. It's very complicated for what is probably very little usage.

PR #338, opened 2024-09-05
react: use stable references to event listeners in hooks

AndyTWF:
### Context

N/A

### Description

A number of our hooks allow users to pass in their own listeners to be able to monitor events such as new messages or occupancy updates, even if the hook exposes the latest data.

At present, we use the listener param in a useEffect to apply listener subscriptions. This causes listener cycling if the hook is called using an inline listener, as the reference changes every time. This behaviour is not desired, as it may cause the subscriber to miss events.

This change fixes the issue by introducing a new internal hook, useEventListenerRef. This hook creates a stable reference to the passed in listener between re-renders. So whilst the underlying listener may change, we avoid cycling subscriptions on every render and thus potentially missing messages.

We also replace all instances of listener use in the other hooks with calls to this new hook, so that the listener passed in is maintained.

### Checklist

* [x] QA'd by the author.

On main, try the in-repo demo app using trace level logging. If you type something into the input, you'll see that the useRoomReactions listener unmounts and remounts.

On this branch, you'll see that no longer happens.

### DIFF
--- a/src/react/README.md
+++ b/src/react/README.md
@@ -54,6 +54,8 @@ At a client-level, changes to the connection status can be subscribed to by prov
 
 Listeners can be provided for events related to a particular `Room` instances status changes via `onRoomStatusChange`. Furthermore, discontinuities in the event stream for room features can be monitored via the `onDiscontinuity` event listener. Room-level events pertain to the nearest parent `RoomProvider` in the React component tree.
 
+Changing the value provided for each of the available listeners will cause the previously registered listener instance to stop receiving events. However, all message events will be received by exactly one listener. If your listener becomes `undefined`, then the subscription will be removed.
+
 ```tsx
 import { useOccupancy } from '@ably/chat/react';
 
@@ -210,13 +212,19 @@ const MyComponent = () => {
 ### Subscribing To Messages
 
 You can provide an optional listener that will receive the messages sent to the room; if provided, the hook will
-automatically subscribe to messages in the room.
+automatically subscribe to messages in the room. As long as a defined value is provided, the subscription will persist
+across renders: making your listener value `undefined` will cause the subscription to be removed until it becomes defined
+again.
 
 Additionally, providing the listener will allow you to access the `getPreviousMessages` method, which can be used to
 fetch previous messages up until the listener was subscribed.
 
-The `getPreviousMessages` method can be useful when recovering from a discontinuity event,
-as it allows you to fetch all the messages that were missed while the listener was not subscribed.
+The `getPreviousMessages` method can be useful when recovering from a discontinuity event, as it allows you to fetch all
+the messages that were missed while the listener was not subscribed. As long as you provide a defined value for the
+listener (and there are no message discontinuities), `getPreviousMessages` will consistently return messages from the
+same point across renders. However, if your listener becomes `undefined`, then the subscription to messages will be
+removed meaning that, if you then re-define your listener, `getPreviousMessages` will now return messages from the
+new subscription point.
 
 ```tsx
 import { useEffect, useState } from 'react';
--- a/src/react/helper/use-event-listener-ref.ts
+++ b/src/react/helper/use-event-listener-ref.ts
@@ -1,0 +1,57 @@
+import { useCallback, useEffect, useRef } from 'react';
+
+/**
+ * The type of a callback function that can be stored in the reference.
+ */
+type Callback<CallbackArguments extends unknown[]> = (...args: CallbackArguments) => void;
+
+/**
+ * A hook that creates a reference to an event listener callback function. It is used to stabilize the reference
+ * across renders, so that listeners don't get unsubscribed and resubscribed on every render when passed in as a prop.
+ *
+ * For example, doing this:
+ *
+ * ```jsx
+ *   export function MySubscription() {
+ *     useHookWithListener(() => {})
+ *
+ *     return <div>My Subscription</div>
+ *   }
+ * ```
+ *
+ * Where the `useHookWithListener` hook is defined as:
+ *
+ * ```jsx
+ *  export function useHookWithListener(listener) {
+ *    const listenerRef = useEventListenerRef(listener);
+ *    useEffect(() => {
+ *      // Use the listenerRef
+ *    }, [listenerRef]);
+ * }
+ * ```
+ *
+ * Will ensure that the listener is not unsubscribed and resubscribed on every render (i.e. the useEffect will not be called
+ * on every render).
+ *
+ * We allow for the callback to be undefined, as callbacks in the majority of our hooks are optional. In this instance we return undefined,
+ * so that subscriptions will be unwound by the useEffect hook that's using them.
+ *
+ * @internal
+ * @template Arguments - The type of arguments accepted by the callback function.
+ * @param callback - The callback function to be stored in the reference.
+ * @returns A static callback function that wraps the provided callback function, or undefined if no callback is provided.
+ */
+export const useEventListenerRef = <Arguments extends unknown[]>(
+  callback?: Callback<Arguments>,
+): Callback<Arguments> | undefined => {
+  const ref = useRef<Callback<Arguments> | undefined>(callback);
+  useEffect(() => {
+    ref.current = callback;
+  });
+
+  const returnVal = useCallback((...args: Arguments) => {
+    ref.current && ref.current(...args);
+  }, []);
+
+  return callback ? returnVal : undefined;
+};
--- a/src/react/hooks/use-chat-connection.ts
+++ b/src/react/hooks/use-chat-connection.ts
@@ -2,6 +2,7 @@ import { Connection, ConnectionLifecycle, ConnectionStatusChange, ConnectionStat
 import { ErrorInfo } from 'ably';
 import { useEffect, useState } from 'react';
 
+import { useEventListenerRef } from '../helper/use-event-listener-ref.js';
 import { useChatClient } from './use-chat-client.js';
 
 /**
@@ -56,6 +57,9 @@ export const useChatConnection = (options?: UseChatConnectionOptions): UseChatCo
     setCurrentStatus(chatClient.connection.status.current);
   }, [chatClient]);
 
+  // Create stable references for the listeners
+  const onStatusChangeRef = useEventListenerRef(options?.onStatusChange);
+
   // Apply the listener to the chatClient's connection status changes to keep the state update across re-renders
   useEffect(() => {
     chatClient.logger.debug('useChatConnection(); applying internal listener');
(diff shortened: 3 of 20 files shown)


src/react/helper/use-event-listener-ref.ts
```diff
@@ -0,0 +1,54 @@
...
+export const useEventListenerRef = <Arguments extends unknown[]>(
+  callback?: Callback<Arguments>,
+): Callback<Arguments> | undefined => {
+  const ref = useRef<Callback<Arguments> | undefined>(callback);
+  useLayoutEffect(() => {
+    ref.current = callback;
+  });
+
+  return useMemo(() => {
+    return ref.current;
```
splindsay-92, line 52: Do we need the useMemo here? I thought refs don't cause updated when used in deps anyway?

src/react/hooks/use-presence.ts
```diff
@@ -75,6 +76,9 @@ export const usePresence = (params?: UsePresenceParams): UsePresenceResponse =>
   // store the roomStatus in a ref to ensure the correct value is used in the effect cleanup
   const roomStatusAndConnectionStatusRef = useRef({ roomStatus, connectionStatus });
 
+  // create a stable reference for the onDiscontinuity listener
+  const onDiscontinuityRef = useEventListenerRef(params?.onDiscontinuity);
```
splindsay-92, line 80: I can't see anywhere in the `typedocs` that specifies the new behaviour of the hooks. I can see it's in the readme, should we make it explicitly clear in each hook too?

src/react/helper/use-event-listener-ref.ts
AndyTWF: Yeah it's probably overkill, removed

src/react/hooks/use-presence.ts
AndyTWF: Yes, I've documented this on the `StatusParams` type.

src/react/README.md
```diff
@@ -54,6 +54,8 @@ At a client-level, changes to the connection status can be subscribed to by prov
 
 Listeners can be provided for events related to a particular `Room` instances status changes via `onRoomStatusChange`. Furthermore, discontinuities in the event stream for room features can be monitored via the `onDiscontinuity` event listener. Room-level events pertain to the nearest parent `RoomProvider` in the React component tree.
 
+Listeners passed in are stored internally to create a stable reference to provide stability to the listener subscription. Therefore, the listener values passed to hooks are **ignored after the first render**. If you have optional functionality in your application, you should handle this within the listener, rather than swapping out listener functions.
```
vladvelici, line 57: Do we mean that if you

```
// first render
useFeature({listener: () => console.log("batman")});

// second render
useFeature({listener: () => console.log("yoda")});
```

this always prints `batman` on events?

I thought line 48, updating the ref.current in useLayoutCallback, makes sure we actually call the correct listener, so outputting `yoda` for events after second render?

splindsay-92 (approved):
Looks good!

src/react/helper/use-event-listener-ref.ts
AndyTWF: On second thought @splindsay-92, we need to have it otherwise the handler gets updated every render - see latest commit :)

src/react/README.md
AndyTWF: Yes that's correct - although we use `useLayoutEffect` we memoize the value on the first render, so we always get the same thing back.

splindsay-92 (approved):
All looks good

src/react/helper/use-event-listener-ref.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+export const useEventListenerRef = <Arguments extends unknown[]>(
+  callback?: Callback<Arguments>,
+): Callback<Arguments> | undefined => {
+  const ref = useRef<Callback<Arguments> | undefined>(callback);
+  useEffect(() => {
+    ref.current = callback;
+  }, [callback]);
+
+  return useMemo(() => ref.current, []);
```
vladvelici, line 51: If this is what we do now, why the indirection and not just

```
return useMemo(() => callback, [])
```

?

AndyTWF: To do this you'd have to pass `callback ` to dependencies array, which in turn would cause `useMemo` to re-run every time and thus you'd get an ever-changing value.

vladvelici: If the function passed to `useMemo` is only called once on first render (which I think it is due to empty deps array), then `callback` or `ref.current` has the same effect there.

Oh -- but if you
```
return useMemo(() => { return (...args) => ref.current(...args) }, [])
```

then it's different I think?

test/react/helper/use-event-listener-ref.test.tsx
vladvelici, line 1: I think a test that also checks which callback gets called would be useful?

Back to my batman/yoda example... we need to make sure the correct user-provided callback is called while the reference is kept constant?

AndyTWF: Added a test to this effect

src/react/helper/use-event-listener-ref.ts
AndyTWF: Yeah you're right - fix incoming!

src/react/hooks/use-chat-connection.ts
```diff
@@ -73,16 +77,15 @@ export const useChatConnection = (options?: UseChatConnectionOptions): UseChatCo
 
   // Register the listener for the user-provided onStatusChange callback
   useEffect(() => {
-    if (!options?.onStatusChange) return;
+    if (!onStatusChangeRef) return;
```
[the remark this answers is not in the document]
vladvelici: We might adjust to pass the `undefined` unmodified (will comment in the new hook)

src/react/helper/use-event-listener-ref.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @internal
+ * @template Arguments - The type of arguments accepted by the callback function.
+ * @param callback - The callback function to be stored in the reference.
+ * @returns The callback function stored in the reference.
+ */
+export const useEventListenerRef = <Arguments extends unknown[]>(
+  callback?: Callback<Arguments>,
+): Callback<Arguments> | undefined => {
+  const ref = useRef<Callback<Arguments> | undefined>(callback);
```
vladvelici, line 46: This never returns undefined now, but maybe it should if it gets called with undefined? To make sure we don't subscribe if not needed and unsubscribe if listener changes to `undefined`.

I guess adding a line
```
if (!callback) { return undefined; }
``` 
somewhere at the top? Not sure if before or after the `useRef` is ideal

src/react/hooks/use-occupancy.ts
```diff
@@ -60,16 +61,20 @@ export const useOccupancy = (params?: UseOccupancyParams): UseOccupancyResponse
...
+  const listenerRef = useEventListenerRef(params?.listener);
+  const onDiscontinuityRef = useEventListenerRef(params?.onDiscontinuity);
+
   // if provided, subscribes the user provided discontinuity listener
   useEffect(() => {
-    if (!params?.onDiscontinuity) return;
+    if (!onDiscontinuityRef) return;
     logger.debug('useOccupancy(); applying onDiscontinuity listener', { roomId: room.roomId });
-    const { off } = room.occupancy.onDiscontinuity(params.onDiscontinuity);
+    const { off } = room.occupancy.onDiscontinuity(onDiscontinuityRef);
```
splindsay-92, line 72: Is this supposed to be onDiscontinuityRef.current?

AndyTWF: No - because under the hood its just a function, not the return type of `React.useRef`

src/react/hooks/use-chat-connection.ts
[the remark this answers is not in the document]
AndyTWF: See comment on new hook :)

src/react/helper/use-event-listener-ref.ts
vladvelici: This was playing in my mind so I gave it a go...

We can't
```
if (bla) {
  useSomething()
}
```
but `useSomething` can return conditionally.

So all that means is that we have to move the condition to the bottom.
```
const stableCallback = useCallback(...) // as now but don't return
if (!callback) return undefined;
return stableCallback;
```

I've tried it here: https://github.com/ably/ably-chat-js/commit/c152f810f7c6e85c4eada3e015c9301e9d378aeb, seems to work

AndyTWF: Modified inline with this after standup discussion

src/react/helper/use-event-listener-ref.ts
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    ref.current = callback;
+  });
+
+  const returnVal = useCallback((...args: Arguments) => {
+    ref.current && ref.current(...args);
+  }, []);
```
vladvelici, line 54: nitpick: Should we pass `ref` in deps here? It never changes so it shouldn't matter, but the rule is to pass all things used inside the callback.

vladvelici (approved):
✅

src/react/helper/use-event-listener-ref.ts
splindsay-92: I think an empty array is a valid option since refs don't cause updates like states do

AndyTWF: Looking around other places, nobody else seems to because ref's are guaranteed constant - so it shouldn't make any difference (just a preference thing) :)

test/react/helper/use-event-listener-ref.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    rerender({ callback: newCallback });
+
+    (result.current as (arg0: string, arg1: string) => void)('arg3', 'arg4');
+    expect(newCallback).toHaveBeenCalledWith('arg3', 'arg4');
+    expect(originalCallback).lastCalledWith('arg1', 'arg2');
+
+    expect(result.current).toBe(initialResult);
+  });
+
+  it('should return handle undefined callbacks ', () => {
```
splindsay-92, line 28: Just a little nitpick, should this be, should handle undefined callbacks?

AndyTWF: Fixed!

test/react/hooks/use-messages.integration.test.tsx
```diff
@@ -188,4 +188,207 @@ describe('useMessages', () => {
...
+    const roomId = randomRoomId();
+    const room = chatClient.rooms.get(roomId, RoomOptionsDefaults);
+
+    // send a few messages before the first room has subscribed
+    await room.messages.send({ text: 'The force is strong with this one' });
+    await room.messages.send({ text: 'I have the high ground' });
+    await room.messages.send({ text: 'You underestimate my power' });
+
+    // Wait 1 second to make sure all messages are received
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
splindsay-92, line 205: Should we avoid arbitrary waits? Also, do we need it? I thought that because the realtime serial we get from attachments must always be after sending of the last few messages (since this happens serially in channel) we wouldn't need to wait beyond sending of the messages above? :)

AndyTWF: We needed a small wait to ensure messages propagate around realtime - not waiting would sometimes mean the re-render would happen a tad too quick :)

splindsay-92 (approved):

PR #339, opened 2024-09-05
react-hooks: updated logger mock in hook tests.

splindsay-92:
### Context

* Some tests were failing due to an infinite loop caused by how the `useLogger` mock was defined in tests. When the hooks re-render, they call the `useLogger` hook, we provided an arrow function that returns a new instance each time, this caused any `useEffect` dependant on it to re-render.

### Description

*  To better mock the expected behaviour of the `useLogger` hook, which should return a memoized object, I have moved the logger creation to the `beforeEach` test setup. This way, it is created only once per test.


test/react/hooks/use-room-reactions.test.tsx
```diff
@@ -7,6 +7,7 @@ import { makeTestLogger } from '../../helper/logger.ts';
 import { makeRandomRoom } from '../../helper/room.ts';
 
 let mockRoom: Room;
+let mockLogger: ReturnType<typeof makeTestLogger>;
```
vladvelici, line 10: Not sure how this works. Is this supposed to have a value?

splindsay-92: The value is set in the `beforeEach` function, so will always be set before each test.

AndyTWF (approved):

PR #342, opened 2024-09-06
react: ensure we don't show old errors in typing indicators

vladvelici:
### Context

I've noticed that this error was shown briefly when loading the demo app and started digging:

Error that flashes when loading demo app

This only happens when you are still waiting for something from a room that's been released, which means we update an old state. The room is only released when a parent component of what we see there (ChatRoomProvider) is unmounted.

This PR adds a new convenience function `unsubscribablePromise` to make it easy to ensure old promises don't settle. IT also clears old errors when the effect runs.

The old version of this code was also not registering the unsubscribe from the hook correctly.

### Checklist

* [x] QA'd by the author.
* [ ] Unit tests created (if applicable).
* [ ] Integration tests created (if applicable).
* [ ] Follow coding style guidelines found [here](https://github.com/ably/engineering/tree/main/best-practices).
* [x] TypeDoc updated (if applicable).
* [ ] (Optional) Update documentation for new features.
* [ ] Browser tests created (if applicable).
* [x] In repo demo app updated (if applicable).

### Testing Instructions (Optional)

Open demo app - enjoy the absence of the flashing error.

PR #344, opened 2024-09-06
react: ensure we don't show old errors in typing indicators

vladvelici:
### Context

I've noticed that this error was shown briefly when loading the demo app and started digging:

Error that flashes when loading demo app

This only happens when you are still waiting for something from a room that's been released, which means we update an old state. The room is only released when a parent component of what we see there (ChatRoomProvider) is unmounted.

The existing version of this code was also not setting the unsubscribe from the hook correctly.

This version uses a simple `mounted` variable to address the problem. Commits in a [previous PR #342] (https://github.com/ably/ably-chat-js/pull/342) attempted 2 different approaches but both turned out be more trouble than worth.

### Checklist

* [x] QA'd by the author.

Open demo app - enjoy the absence of the flashing error.

### DIFF
--- a/src/react/hooks/use-typing.ts
+++ b/src/react/hooks/use-typing.ts
@@ -1,6 +1,6 @@
-import { Typing, TypingEvent, TypingListener, TypingSubscriptionResponse } from '@ably/chat';
+import { Typing, TypingEvent, TypingListener } from '@ably/chat';
 import * as Ably from 'ably';
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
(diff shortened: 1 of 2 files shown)


cspell.json
```diff
@@ -15,7 +15,8 @@
     "Suspendable",
     "Failable",
     "livestreams",
-    "livestream"
+    "livestream",
+    "unsubscribable"
```
splindsay-92, line 19: I can't see where this is used in the PR, is it still needed?

splindsay-92: I've removed as I don't think it was needed.

splindsay-92 (approved):
Looks good, thanks Vlad!

cspell.json
vladvelici: i forgot it, thanks!

PR #352, opened 2024-09-18
React unhandled rejection errors

### LINKED ISSUE #350: "Error: Room released before initialization started" when using "@ably/chat/react" providers in my nextjs app.
I have the following layout.tsx:

```tsx
// layout.tsx
import { LobbyRoomProvider } from "@/app/[locale]/(dynamic)/cs/lobby/[room]/_components/lobby-room-provider"
import { getAuthUser } from "@repo/supabase/queries"

interface LobbyLayoutProps {
  children: React.ReactNode
  params: {
    room: string
  }
}
export default async function LobbyLayout({
  children,
  params,
}: LobbyLayoutProps) {
  const { room } = params
  const { data } = await getAuthUser()
  const user = data?.user
  return (
    <LobbyRoomProvider initialRoom={room} initialUser={user}>
      {children}
    </LobbyRoomProvider>
  )
}
```

I then have the following client component which has the ably providers and so on:
```tsx
// lobby-room-provider.tsx
"use client"
import * as Ably from "ably"
import { ChatClientProvider, ChatRoomProvider } from "@ably/chat/react"
import { ChatClient, LogLevel, RoomOptionsDefaults } from "@ably/chat"
import { useMemo } from "react"
import { useUser } from "@supabase/auth-helpers-react"
import type { User } from "@supabase/supabase-js"

interface LobbyRoomProviderProps {
  children: React.ReactNode
  initialUser: User | null
  initialRoom: string
}
export function LobbyRoomProvider(props: LobbyRoomProviderProps) {
  const { children, initialUser, initialRoom } = props
  const clientUser = useUser()
  const user = clientUser ?? initialUser
  const ablyChatRoom = `lobby:${initialRoom}`

  const isLoggedIn = !!user

  const client = useMemo(() => {
    return new Ably.Realtime({
      authUrl: isLoggedIn ? "/api/ably" : undefined,
      autoConnect: typeof window !== "undefined",
      clientId: isLoggedIn ? user.id : undefined,
      key: process.env.NEXT_PUBLIC_ABLY_API_KEY,
      logHandler: console.log,
      logLevel: 2,
    })
  }, [user, isLoggedIn])

  const chatClient = useMemo(() => {
    return new ChatClient(client, {
      logLevel: LogLevel.Info,
      logHandler: console.log,
    })
  }, [client])

  return (
    <ChatClientProvider client={chatClient}>
      <ChatRoomProvider id={ablyChatRoom} options={RoomOptionsDefaults}>
        {children}
      </ChatRoomProvider>
    </ChatClientProvider>
  )
}
```

Loading the app gives me this error immediately:
CleanShot 2024-09-17 at 19 08 36@2x

My logs also show the following error:

```
Ably: Channel state for channel "lobby:1::$chat::$reactions": detached; reason: [_ErrorInfo: Channel detached; statusCode=404; code=90001] 2

```

I was under the impression that it would create the channel for me if it does not exist. Is that not correct?

AndyTWF:
### Context

Fixes #350 

### Description

- Adds a promise rejection handler in ChatRoomProvider to handle unhandled rejections (due to initialization skips)
- Adds a promise rejection handler for getPreviousMessages if the room is released before initialization
- Changes the room release before initialization to use an appropriate error code
- Avoid error logging in usePresenceListener and useTyping if room is released

### Checklist

* [x] QA'd by the author.

Open demo app, refresh many times - notice there are no errors (compare to main).

### DIFF
--- a/src/core/errors.ts
+++ b/src/core/errors.ts
@@ -1,3 +1,5 @@
+import * as Ably from 'ably';
+
 /**
  * Error codes for the Chat SDK.
  */
@@ -86,3 +88,13 @@ export enum ErrorCodes {
    */
   RoomLifecycleError = 102105,
 }
+
+/**
+ * Returns true if the {@link Ably.ErrorInfo} code matches the provided ErrorCodes value.
+ *
+ * @param errorInfo The error info to check.
+ * @param error The error code to compare against.
+ * @returns true if the error code matches, false otherwise.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+export const errorInfoIs = (errorInfo: Ably.ErrorInfo, error: ErrorCodes): boolean => errorInfo.code === error;
--- a/src/core/room.ts
+++ b/src/core/room.ts
@@ -2,6 +2,7 @@ import * as Ably from 'ably';
 import cloneDeep from 'lodash.clonedeep';
 
 import { ChatApi } from './chat-api.js';
+import { ErrorCodes } from './errors.js';
 import { Logger } from './logger.js';
 import { DefaultMessages, Messages } from './messages.js';
 import { DefaultOccupancy, Occupancy } from './occupancy.js';
@@ -186,11 +187,13 @@ export class DefaultRoom implements Room {
     this._status = new DefaultStatus(logger);
 
     // Room initialization: handle the state before features start to be initialized
+    // We don't need to catch it here as we immediately go into _asyncOpsAfter where we catch it
     const rejectablePromise = makeRejectablePromise(makeAlwaysResolves(initAfter));
     const initFeaturesAfter = rejectablePromise.promise;
+
     this._finalizer = () => {
       const rejectedNow = rejectablePromise.reject(
-        new Ably.ErrorInfo('Room released before initialization started.', 40000, 400),
+        new Ably.ErrorInfo('Room released before initialization started', ErrorCodes.RoomIsReleased, 400),
       );
       if (rejectedNow) {
         // when this promise is rejected by calling reject(), the room is released
@@ -198,14 +201,6 @@ export class DefaultRoom implements Room {
       }
       return initAfter;
     };
-
-    // Catch to prevent global errors and print a debug log
-    initFeaturesAfter.catch((error: unknown) => {
-      this._logger.debug('Room initialization was prevented before initializing features', {
-        error: error,
-        roomId: roomId,
-      });
-    });
 
     // Setup features
     this._messages = new DefaultMessages(
@@ -260,6 +255,8 @@ export class DefaultRoom implements Room {
     this._asyncOpsAfter = this._setupAsyncRoomInit(features, initFeaturesAfter, realtime);
 
     // Catch errors from asyncOpsAfter to prevent unhandled promise rejection error and print a debug log
+    // This only prevents the base promise from erroring - calls to attach, detach etc that depend on _asyncOpsAfter
+    // will need to specify their own error handling.
     this._asyncOpsAfter.catch((error: unknown) => {
       this._logger.debug('Room initialization was prevented before finishing', { error: error, roomId: roomId });
     });
@@ -298,7 +295,7 @@ export class DefaultRoom implements Room {
         this._finalizer = () => {
           // Make sure async ops (attach or detach) don't run after calling release()
           rejectableAsyncOpsAfter.reject(
-            new Ably.ErrorInfo('Room released before initialization finished', 40000, 400),
+            new Ably.ErrorInfo('Room released before initialization finished', ErrorCodes.RoomIsReleased, 400),
           );
           return finalizerFuncPromise.then((f) => {
             return f();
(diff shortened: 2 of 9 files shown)


src/core/errors.ts
```diff
@@ -86,3 +88,6 @@ export enum ErrorCodes {
    */
   RoomLifecycleError = 102105,
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+export const errorInfoIs = (errorInfo: ErrorInfo, error: ErrorCodes): boolean => errorInfo.code === error;
```
splindsay-92, line 93: Should we typedocs this as it's exported and used elsewhere?

src/core/room.ts
```diff
@@ -188,9 +189,19 @@ export class DefaultRoom implements Room {
     // Room initialization: handle the state before features start to be initialized
     const rejectablePromise = makeRejectablePromise(makeAlwaysResolves(initAfter));
     const initFeaturesAfter = rejectablePromise.promise;
+    // Catch to prevent global errors and print a debug log
```
vladvelici, line 192: This one might not be needed because we always create `_asyncOpsAfter`, which is using `initFeaturesAfter.then(...)` and it is created sync in this constructor, so before `initFeaturesAfter` can be rejected.

`_asyncOpsAfter` has a catch already, at the bottom of the constructor.

src/core/errors.ts
AndyTWF: Ah my bad, missed this one - fixed!

src/react/hooks/use-presence-listener.ts
```diff
@@ -201,7 +201,13 @@ export const usePresenceListener = (params?: UsePresenceListenerParams): UsePres
...
       .catch((error: unknown) => {
-        logger.error('usePresenceListener(); error fetching initial presence data', { error, roomId: room.roomId });
+        const errorInfo = error as Ably.ErrorInfo;
+        if (errorInfoIs(errorInfo, ErrorCodes.RoomIsReleased)) return;
+
+        logger.error('usePresenceListener(); error fetching initial presence data', {
+          error,
+          roomId: room.roomId,
+        });
         setErrorState(error as Ably.ErrorInfo);
```
vladvelici, line 211: nitpick: use the `errorInfo` var here to avoid casting again?

test/core/messages.integration.test.ts
```diff
@@ -441,4 +443,17 @@ describe('messages integration', () => {
...
+    const chat = context.chat;
+    const roomId = randomRoomId();
+    const room = chat.rooms.get(roomId, RoomOptionsDefaults);
+
+    // Create a subscription to messages
+    room.messages.subscribe(() => {});
+
+    // Now release the room
+    // We should not have any unhanded promise rejections
+    await chat.rooms.release(roomId);
```
vladvelici, line 457: Is this test working correctly if there are unhandled promise rejections? Those errors get thrown outside of the normal execution, so they don't typically make anything crash.

Not sure how to listen to them exactly - maybe listening to https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event?

test/core/room.test.ts
```diff
@@ -514,4 +515,11 @@ describe('Room', () => {
     expect(attached).toBe(false);
     expect(attachedError).toBe(true);
   });
+
+  it<TestContext>('can be released immediately without unhandled rejections', async (context) => {
+    const room = context.getRoom(defaultRoomOptions);
+
+    // Release the room
+    await (room as DefaultRoom).release();
```
vladvelici, line 523: same as before - does this test actually fail if there are unhandled rejections?

splindsay-92 (approved):
LGTM!

test/core/messages.integration.test.ts
AndyTWF: The test will still "pass" - unfortunately we can't get around that. But the actual test process will exit with code `1` and will therefore fail CI.

I've added a comment to this effect.

test/core/room.test.ts
AndyTWF: As above

src/react/hooks/use-presence-listener.ts
AndyTWF: Done!

src/core/room.ts
AndyTWF: We'll still need it.

The `initFeaturesAfter.then` and `initFeaturesAfter.catch` calls returns a new promise that executes based on the result of `initFeaturesAfter`.

So the call to `initFeaturesAfter.catch` here is one promise chain, `_asyncOpsAfter` is another separate one. Both of them will reject if `_initFeaturesAfter` rejects.

If we wanted to not have this behaviour, then `initFeaturesAfter` would have to be set to the value of `initFeaturesAfter.catch(...)` - which I don't think is desirable.

AndyTWF: Yeah - I've removed it and left a comment to that effect.

vladvelici (approved):

PR #362, opened 2024-10-08
materialization: add fields to support deletion and edits in chat.

splindsay-92:
### Context

* As part of the support work for materialization, changes to the DefaultMessage class are needed. Primarily, this involves adding new fields to support deletes and edits, as well as moving to the new `message.actions` field.
* With the addition of `message.actions`, we will now use this to filter chat messages, the old `message.name` will still be used to filter things like `[meta]occupancy`

### Description

* Adds the new fields to the DefaultMessage class.
* Updates the message parser to handle the new fields.
* Adds the support for the new `message.name` name filter.

### Checklist

* [x] QA'd by the author.

## Summary by CodeRabbit

- **New Features**
	- Enhanced message handling with new properties and actions for chat messages.
	- Added support for updated and deleted message events.
	- Introduced new enumerations for message actions and types.
	- New type definition for action metadata associated with chat messages.

- **Bug Fixes**
	- Improved error handling in event emitters for better validation.

- **Tests**
	- Expanded test coverage for message actions and event subscriptions.
	- Updated tests to reflect the new message structure and properties.

- **Documentation**
	- Clarified changes in the test cases to align with new message formats.

### DIFF
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -4,6 +4,46 @@
 export enum MessageEvents {
   /** Fires when a new chat message is received. */
   Created = 'message.created',
+
+  /** Fires when a chat message is updated. */
+  Updated = 'message.updated',
+
+  /** Fires when a chat message is deleted. */
+  Deleted = 'message.deleted',
+}
+
+/**
+ * Realtime chat message names.
+ */
+export enum RealtimeMessageNames {
+  /** Represents a regular chat message. */
+  ChatMessage = 'chat.message',
+}
+
+/**
+ * Chat Message Actions.
+ */
+export enum ChatMessageActions {
+  /** Represents a message with no action set. */
+  MessageUnset = 'message.unset',
+
+  /** Action applied to a new message. */
+  MessageCreate = 'message.create',
+
+  /** Action applied to an updated message. */
+  MessageUpdate = 'message.update',
+
+  /** Action applied to a deleted message. */
+  MessageDelete = 'message.delete',
+
+  /** Action applied to a new annotation. */
+  MessageAnnotationCreate = 'annotation.create',
+
+  /** Action applied to a deleted annotation. */
+  MessageAnnotationUpdate = 'annotation.delete',
+
+  /** Action applied to a meta occupancy message. */
+  MessageMetaOccupancy = 'meta.occupancy',
 }
 
 /**
--- a/src/core/helpers.ts
+++ b/src/core/helpers.ts
@@ -1,6 +1,6 @@
 import * as Ably from 'ably';
 
-import { MessageEvents, RoomReactionEvents } from './events.js';
+import { RealtimeMessageNames, RoomReactionEvents } from './events.js';
 import { Message } from './message.js';
 import { parseMessage } from './message-parser.js';
 import { Reaction } from './reaction.js';
@@ -89,7 +89,7 @@ export async function reactionFromEncoded(encoded: unknown): Promise<Reaction> {
  */
 export function getEntityTypeFromAblyMessage(message: Ably.InboundMessage): ChatEntityType {
   switch (message.name) {
-    case MessageEvents.Created: {
+    case RealtimeMessageNames.ChatMessage: {
       return ChatEntityType.ChatMessage;
     }
     case RoomReactionEvents.Reaction: {
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -2,6 +2,7 @@
  * @module chat-js
  */
 
+export type { ActionMetadata } from './action-metadata.js';
 export { ChatClient } from './chat.js';
 export type { ClientOptions } from './config.js';
 export type {
@@ -13,7 +14,7 @@ export type {
 export { ConnectionStatus } from './connection.js';
 export type { DiscontinuityListener, OnDiscontinuitySubscriptionResponse } from './discontinuity.js';
 export { ErrorCodes, errorInfoIs } from './errors.js';
-export { MessageEvents, PresenceEvents } from './events.js';
+export { ChatMessageActions, MessageEvents, PresenceEvents } from './events.js';
 export type { Headers } from './headers.js';
 export {
   ChatEntityType,
@@ -26,7 +27,7 @@ export {
 } from './helpers.js';
 export type { LogContext, Logger, LogHandler } from './logger.js';
 export { LogLevel } from './logger.js';
-export type { Message, MessageHeaders, MessageMetadata } from './message.js';
+export type { Message, MessageActionDetails, MessageHeaders, MessageMetadata } from './message.js';
 export type {
   MessageEventPayload,
   MessageListener,
--- a/src/core/message-parser.ts
+++ b/src/core/message-parser.ts
@@ -1,6 +1,7 @@
 import * as Ably from 'ably';
 
-import { DefaultMessage, Message, MessageHeaders, MessageMetadata } from './message.js';
+import { ChatMessageActions } from './events.js';
+import { DefaultMessage, Message, MessageActionDetails, MessageHeaders, MessageMetadata } from './message.js';
 
 interface MessagePayload {
   data?: {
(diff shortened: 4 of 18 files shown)


src/core/message.ts
```diff
@@ -70,6 +75,57 @@ export interface Message {
...
+
+  /**
+   * The clientId of the user who edited the message.
+   * If the message has not been edited, or has been edited by a connection without a clientId (such as requests made
+   * using an API key only), this value is undefined.
+   */
+  readonly editedBy?: string;
+
+  /**
+   * The detail of the edit. If the message has not been edited, this value is undefined.
```
vladvelici, line 112:
```suggestion
   * The detail of the latest edit. If the message has not been edited, this value is undefined.
```

vladvelici (approved):
LGTM 👍

Good to merge after tests pass

src/core/message-parser.ts
```diff
@@ -10,9 +10,30 @@ interface MessagePayload {
...
   extras?: {
-    timeserial?: string;
     headers?: MessageHeaders;
   };
+
+  serial: string;
+  updatedAt?: number;
+  deletedAt?: number;
+  action: Ably.MessageAction;
+  operation?: Ably.Operation;
```
vladvelici, line 20: What's `operation` and what's `action`?

src/core/events.ts
```diff
@@ -4,6 +4,23 @@
 export enum MessageEvents {
   /** Fires when a new chat message is received. */
   Created = 'message.created',
+
+  /** Fires when a chat message is updated. */
+  Edited = 'message.edited',
```
vladvelici, line 9:
```suggestion
  Updated = 'message.updated',
```
?

src/core/messages.ts
```diff
@@ -26,6 +26,26 @@ import EventEmitter from './utils/event-emitter.js';
  */
 interface MessageEventsMap {
   [MessageEvents.Created]: MessageEventPayload;
+  [MessageEvents.Edited]: MessageEventPayload;
```
vladvelici, line 29:
```suggestion
  [MessageEvents.Updated]: MessageEventPayload;
```
? Not sure if we're going with `update` everywhere?

test/core/message-parser.test.ts
```diff
@@ -45,22 +55,75 @@ describe('parseMessage', () => {
...
+        clientId: 'client1',
+        timestamp: 1234567890,
+        extras: {},
+        serial: 'cbfkKvEYgBhDaZ38195418@1728402074206-0:0',
+        action: 'UNHANDLED_ACTION',
+      },
+      expectedError: 'received incoming message with unhandled action; UNHANDLED_ACTION',
+    },
+    {
+      description: 'message.editedAt is undefined for update',
```
vladvelici, line 102:
```suggestion
      description: 'message.updatedAt is undefined for update',
```

src/core/message-parser.ts
vladvelici: I figured it out from the ably-js PR.

src/core/message-parser.ts
```diff
@@ -41,17 +62,71 @@ export function parseMessage(roomId: string | undefined, message: Ably.InboundMe
     throw new Ably.ErrorInfo(`received incoming message without extras`, 50000, 500);
   }
 
-  if (!messageCreatedMessage.extras.timeserial) {
+  if (!messageCreatedMessage.serial) {
```
vladvelici, line 65: Can we rename `messageCreatedMessage` variable since it's no longer just new messages that get here?

src/core/message-parser.ts
```diff
@@ -41,17 +62,71 @@ export function parseMessage(roomId: string | undefined, message: Ably.InboundMe
...
-    messageCreatedMessage.data.metadata ?? {},
-    messageCreatedMessage.extras.headers ?? {},
+    text: messageCreatedMessage.data.text,
+    createdAt: new Date(messageCreatedMessage.timestamp),
+    metadata: messageCreatedMessage.data.metadata ?? {},
+    headers: messageCreatedMessage.extras.headers ?? {},
+  };
+
+  switch (messageCreatedMessage.action) {
+    case 'MESSAGE_CREATE': {
```
vladvelici, line 88: Should this use the [Ably.MessageAction[s]](https://github.com/ably/ably-js/pull/1888/files/2a59fb45d8243f022677143af551217e71eefaf9#r1796692076) thing?

splindsay-92: I'm not sure what you mean, actions is defined above as `Ably.MessageAction` already?

src/core/messages.ts
splindsay-92: Yeh we should, I've made the change

src/core/events.ts
```diff
@@ -4,6 +4,25 @@
 export enum MessageEvents {
   /** Fires when a new chat message is received. */
   Created = 'message.created',
+
+  /** Fires when a chat message is updated. */
+  Updated = 'message.updated',
+
+  /** Fires when a chat message is deleted. */
+  Deleted = 'message.deleted',
```
AndyTWF, line 12: If we support hard deleting in the future, do we need to differentiate here?

src/core/index.ts
```diff
@@ -27,7 +27,7 @@ export {
 } from './helpers.js';
 export type { LogContext, Logger, LogHandler } from './logger.js';
 export { LogLevel } from './logger.js';
-export type { Message, MessageHeaders, MessageMetadata } from './message.js';
+export type { Message, MessageDetails, MessageDetailsMetadata,MessageHeaders, MessageMetadata } from './message.js';
```
AndyTWF, line 30: Think we need to run `npm run format` here

src/core/message-parser.ts
```diff
@@ -10,48 +10,119 @@ interface MessagePayload {
...
     throw new Ably.ErrorInfo(`received incoming message without text`, 50000, 500);
   }
 
-  if (!messageCreatedMessage.extras) {
+  if (!message.extras) {
     throw new Ably.ErrorInfo(`received incoming message without extras`, 50000, 500);
   }
 
-  if (!messageCreatedMessage.extras.timeserial) {
+  if (!message.serial) {
```
AndyTWF, line 65: Worth updating this message to say `serial` rather than `timeserial`?

src/core/message-parser.ts
```diff
@@ -10,48 +10,119 @@ interface MessagePayload {
...
     throw new Ably.ErrorInfo(`received incoming message without timeserial`, 50000, 500);
   }
 
-  return new DefaultMessage(
-    messageCreatedMessage.extras.timeserial,
-    messageCreatedMessage.clientId,
+  let operationDetails: MessageDetails | undefined;
```
AndyTWF, line 69: Do we need any validation around the operation?

src/core/message-parser.ts
```diff
@@ -10,48 +10,119 @@ interface MessagePayload {
...
+    }
+    case 'MESSAGE_DELETE': {
+      if (!message.deletedAt) {
+        throw new Ably.ErrorInfo(`received incoming deletion message without deletedAt`, 50000, 500);
+      }
+      newMessage.deletedBy = message.operation?.clientId;
+      newMessage.deletionDetail = operationDetails;
+      break;
+    }
+    default: {
```
AndyTWF, line 109: I think we should move the `message.action` check to the top and simply ignore messages that don't have a known action.

There's been a push recently to make it so that SDKs do not complain if "new" fields are used - if we see an action that we don't understand, we should just ignore it (perhaps debug log it)

src/core/message.ts
```diff
@@ -115,13 +185,27 @@ export class DefaultMessage implements Message {
     public readonly createdAt: Date,
     public readonly metadata: MessageMetadata,
     public readonly headers: MessageHeaders,
+    public readonly deletedAt?: Date,
```
AndyTWF, line 188: One for you and @vladvelici to discuss where it belongs - we need to do the CHADR-064 update to timeserials here.

src/core/messages.ts
```diff
@@ -26,6 +26,26 @@ import EventEmitter from './utils/event-emitter.js';
...
 interface MessageEventsMap {
   [MessageEvents.Created]: MessageEventPayload;
+  [MessageEvents.Updated]: MessageEventPayload;
+  [MessageEvents.Deleted]: MessageEventPayload;
+}
+
+// This function is used to convert the message action from Ably to the ChatMessageEvents enum
+export function messageActionToMessageEvent(action: Ably.MessageAction): MessageEvents {
+  switch (action) {
+    case 'MESSAGE_CREATE': {
```
AndyTWF, line 36: Can we make these enums, save us repeating ourselves?

src/core/messages.ts
```diff
@@ -479,7 +499,7 @@ export class DefaultMessages
    */
   subscribe(listener: MessageListener): MessageSubscriptionResponse {
```
AndyTWF, line 500: An API question here - should we be allowing users to provide events that they want to subscribe to, a-la presence?

src/core/events.ts
splindsay-92: I'm not too sure, and think that will depend on how it gets implemented in Ably. If it looks the same as a regular deletion message, just with the string cleansed etc.. my gut says no, that we should handle that within the message as part of the operation field, maybe adding it to the metadata object that gets returned, or as a totally separate ably specific field.

src/core/message-parser.ts
splindsay-92: I don't think so, these are coming from realtime, so should be handle by ably-js. As to if they are present, both fields are optional, so we don't need to check that either.

src/core/message-parser.ts
splindsay-92: I would tend to disagree, we are assuming by this stage that you are providing us with a message that fits the schema, if not, we should error. However, we have a check in messages to ignore messages with action types that we don't currently support.

src/core/messages.ts
splindsay-92: I have added the following
```ts
/**
* Enum that maps message actions to message events.
  */
enum MessageActionsToEvents {
  'MESSAGE_CREATE' = MessageEvents.Created,
  'MESSAGE_UPDATE' = MessageEvents.Updated,
  'MESSAGE_DELETE' = MessageEvents.Deleted,
  }
```

src/core/messages.ts
AndyTWF: > Ah yeh, I think we should, I'll make the change to support this as part of this PR.

Doesn't have to be as part of this PR, but would be good to get a DR around it nonetheless - I think there's a discussion to be had around whether we do optional first args (I really dislike the optional first arg pattern!) or whether we do an options object.

src/core/message-parser.ts
AndyTWF: Sure ok - though I think it should only `trace` or `debug` log. Warnings are things that indicate an issue, and us receiving an event we don't know how to handle, is more likely to be non-issue imo

src/core/messages.ts
```diff
@@ -261,7 +282,7 @@ export class DefaultMessages
 
     addListenerToChannelWithoutAttach({
       listener: this._processEvent.bind(this),
-      events: [MessageEvents.Created],
+      events: [RealtimeMessageTypes.ChatMessage, RealtimeMessageTypes.LegacyChatMessage],
```
AndyTWF, line 285: Bitten by our own CI :D We'll need to eslint exclude this.

src/core/messages.ts
```diff
@@ -477,32 +498,59 @@ export class DefaultMessages
   /**
    * @inheritdoc Messages
    */
-  subscribe(listener: MessageListener): MessageSubscriptionResponse {
-    this._logger.trace('Messages.subscribe();');
-    super.on([MessageEvents.Created], listener);
+  subscribe(
```
AndyTWF, line 501: I think there needs to be a DR before we implement filtered events - this is a breaking API change and personally I'm not convinced by optional first arguments.

src/core/message-parser.ts
vladvelici: I meant should it be `Ably.MessageActions.MESSAGE_CREATE` instead of directly a string?

src/core/message.ts
vladvelici: I think we can make a PR for that separately and merge it as soon as we get it done?

src/core/message-parser.ts
splindsay-92: This doesn't work owing to the fact that we haven't exposed this as an enum. But, I have created an enum type we can use, just like we do with Presence :)

src/core/message-parser.ts
splindsay-92: Okay, I've made it so the filter will only emit a debug log if the event is not something we handle yet.

src/core/messages.ts
splindsay-92: Removed as we have bumped the endpoints to use the new type now :)

src/core/messages.ts
splindsay-92: Removed, add a ticket to form a DR for this.

src/core/helpers.ts
```diff
@@ -89,7 +89,7 @@ export async function reactionFromEncoded(encoded: unknown): Promise<Reaction> {
  */
 export function getEntityTypeFromAblyMessage(message: Ably.InboundMessage): ChatEntityType {
   switch (message.name) {
-    case MessageEvents.Created: {
+    case RealtimeMessageTypes.ChatMessage: {
```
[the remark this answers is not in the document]
splindsay-92: If this were not in beta, I would say sure, but as we are and have bumped the endpoints in this change, I think dropping support is fine. If you are testing out with a new version of the SDK, it's likely you don't care about testing against older versions..

src/core/messages.ts
```diff
@@ -26,6 +26,8 @@ import EventEmitter from './utils/event-emitter.js';
  */
 interface MessageEventsMap {
   [MessageEvents.Created]: MessageEventPayload;
+  [MessageEvents.Updated]: MessageEventPayload;
```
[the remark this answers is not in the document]
splindsay-92: Sure, I'll add that in :)

src/core/events.ts
```diff
@@ -4,6 +4,43 @@
...
+ */
+export enum RealtimeMessageTypes {
+  /** Represents a regular chat message. */
+  ChatMessage = 'chat.message',
+}
+
+/**
+ * Chat Message Actions.
+ */
+export enum ChatMessageActions {
```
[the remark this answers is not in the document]
splindsay-92: For the same reasons we do this with presence below, it exports MessageAction events the same way it does for PresenceEvents, but not as an enum type.

src/core/helpers.ts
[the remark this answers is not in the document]
vladvelici: Makes sense

src/core/messages.ts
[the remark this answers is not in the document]
vladvelici: I agree. We could extend the existing interface without actually changing types of any listeners, but we're better off postponing this until it's actually a useful change.

src/core/events.ts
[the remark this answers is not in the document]
vladvelici: Makes sense

vladvelici (approved):
Only thing left from me is https://github.com/ably/ably-chat-js/pull/362#discussion_r1814647355

I'm approving now ✅

src/core/message.ts
```diff
@@ -12,6 +15,29 @@ export type MessageHeaders = Headers;
  */
 export type MessageMetadata = Metadata;
 
+/**
+ * {@link DetailsMetadata} type for a chat messages {@link ActionDetails}.
+ */
+export type ActionDetailsMetadata = DetailsMetadata;
```
vladvelici, line 21: Is DetailsMetadata still used anywhere else? If not, maybe just keep one type?

src/core/message.ts
```diff
@@ -12,6 +15,29 @@ export type MessageHeaders = Headers;
  */
 export type MessageMetadata = Metadata;
 
+/**
+ * {@link DetailsMetadata} type for a chat messages {@link ActionDetails}.
```
vladvelici, line 19:
```suggestion
 * {@link DetailsMetadata} type for a chat message {@link ActionDetails}.
```

src/core/message.ts
```diff
@@ -115,22 +210,66 @@ export class DefaultMessage implements Message {
...
+  get isUpdated(): boolean {
+    return this.updatedAt !== undefined;
+  }
+
+  get isDeleted(): boolean {
+    return this.deletedAt !== undefined;
+  }
+
+  get actionedBy(): string | undefined {
+    return this.latestActionDetails?.clientId;
```
vladvelici, line 237: if action is create should this return the original author as a convenience?

src/core/events.ts
splindsay-92: Considering we aren't quite sure what a hard delete will look like, I'd say we stick with what we have for now.

src/core/message.ts
splindsay-92: It's used in the parser, so I've left it in for now :)

src/core/message.ts
splindsay-92: This has been removed anyway in favour of deletedBy and updatedBy.

vladvelici (approved):

PR #389, opened 2024-11-06
refactored rooms impl

sacOO7:
Easy impl for `rooms.get` and `rooms.release`


sacOO7:
Closing the PR, do let me know if PR helps or if we are on the common ground in terms of understanding the spec.

PR #402, opened 2024-11-14
Refactor message docs for deletions.

splindsay-92:
Refactored the deletion aspect of chat in the README.md, so it keeps consistency with updates and provides greater clarity for its usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,21 +346,66 @@ These additional parameters are:
 * `metadata`: a map of extra information that can be attached to the deletion message.
 
 The return of this call will be the deleted message, as it would appear to other subscribers of the room.
-This is a _soft delete_ and the message will still be available in the history, but with the `deletedAt` property set.
+This is a _soft delete_ and the message will still be available in the history.
 
+Example
 ```ts
-const deletedMessage = await room.messages.delete(message, { description: 'This message was deleted for ...' });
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -346,22 +346,62 @@ These additional parameters are:
...
 ```ts
 const { unsubscribe } = room.messages.subscribe((msg) => console.log(msg));
 ```
 
+#### Handling deletes in realtime
+
+Deletion messages received from realtime have the `latestAction` parameter set to `ChatMessageActions.MessageDelete`, and the event received has the `type` set to `MessageEvents.Deleted`.
+Just like `updates`, `deletion` messages are also full copies of the message, meaning that all that is needed to keep a state or UI up to date is to replace the old message with the received one.
+
+On rare occasions, deletes and updates might arrive over realtime out of order.
```
AndyTWF, line 374: As we provide order guarantees, I think its important to clarify here what "out of order" means. What you mean to say, I think, is that the order messages are received in realtime does not necessarily reflect the global ordering of events (e.g. due to concurrent operations in multiple regions)?

splindsay-92: I've updated the comment here to detail this and also recommend that users should always use the `actionBefore` etc.. for ordering actions.

AndyTWF (approved):

PR #408, opened 2024-11-19
core: lexicographical timeserial comparisons

AndyTWF:
### Context

[CHA-704]

Spec PR: https://github.com/ably/specification/pull/231

### Description

This change removes the parsing of timeserial and instead reverts to comparing them as lexicographically comparable strings. We are therefore able to handle changes in timeserial format from the server without breaking anything.

The timeserial parsing code has thus been removed.

A side-effect is that we can no longer check on message subscriptions, whether getPreviousMessages' end time would be after the subscription point (specification CHA-M5j). This spec point will be removed.

### Checklist

* [x] QA'd by the author.

N/A

[CHA-704]: https://ably.atlassian.net/browse/CHA-704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cspell.json
```diff
@@ -16,6 +16,6 @@
     "livestream",
     "clonedeep"
   ],
-  "ignoreRegExpList": ["/.*@\\d{13}-/"],
+  "ignoreRegExpList": ["/.*@\\d{13}-/", "/\\d{14}-\\d{3}@.*/"],
```
splindsay-92, line 19: Do we still need the previous ignore regex? Also, could you add a comment detailing it will ignore lexicos? :) It'll make it easier at a glance to know what the regex is doing

splindsay-92 (approved):
Just one comment, but it's small so I will approve anyway :)

vladvelici (approved):

cspell.json
AndyTWF: Done and done!

AndyTWF:
Note: this is ready to go, just waiting for a rollout in Sandbox

PR #412, opened 2024-11-20
Room lifecycle in progress flag

### LINKED ISSUE #406: [RoomLifeCycleManager] `_operationInProgress` flag incorrectly set
- Currently, `operationInProgress` flag is [set to false only when `attach` operation finishes.](https://github.com/ably/ably-chat-js/blob/533c4d0314bcce942a30b4d7d6a012c3b081420d/src/core/room-lifecycle-manager.ts#L592) 
- It should be set `false` at the end of other operations like `Detach`, `Release`, and `Internal` operations.
- Also, not sure if we really need this code snippet in place 
https://github.com/ably/ably-chat-js/blob/533c4d0314bcce942a30b4d7d6a012c3b081420d/src/core/room-lifecycle-manager.ts#L130-L134
- This sets `operationInProgress` to `true` when the room has just initialized and not performing any operation.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/CHA-723) by [Unito](https://www.unito.io)

AndyTWF:
### Context

Closes #406 
[CHA-723]

### Description

We weren't properly setting and unsetting the room operation in progress flag (we were only setting it to false when attaching). This change rectifies this.

Also included is the fix for async-mutex operation precedence - as the order was the wrong way around.

### Checklist

* [x] QA'd by the author.

N/A

[CHA-723]: https://ably.atlassian.net/browse/CHA-723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/room-lifecycle-manager.ts
+++ b/src/core/room-lifecycle-manager.ts
@@ -31,11 +31,13 @@ export interface ContributesToRoomLifecycle extends HandlesDiscontinuity {
 /**
  * The order of precedence for lifecycle operations, passed to the mutex which allows
  * us to ensure that internal operations take precedence over user-driven operations.
+ *
+ * The higher the number, the higher the priority.
  */
 enum LifecycleOperationPrecedence {
-  Internal = 0,
+  Internal = 2,
   Release = 1,
-  AttachOrDetach = 2,
+  AttachOrDetach = 0,
 }
 
 /**
--- a/src/core/room-lifecycle-manager.ts
+++ b/src/core/room-lifecycle-manager.ts
@@ -127,12 +127,6 @@ export class RoomLifecycleManager {
     this._transientDetachTimeouts = new Map();
     this._lifecycle = lifecycle;
 
-    // This shouldn't be the case except in testing, but if we're already attached, then we should consider
-    // ourselves not in the middle of an operation and thus consider channel events.
-    if (this._lifecycle.status !== RoomStatus.Attached) {
-      this._operationInProgress = true;
-    }
-
     this._setupContributorListeners(transientDetachTimeout);
   }
 
@@ -229,7 +223,7 @@ export class RoomLifecycleManager {
               channel: contributor.channel.name,
             });
             this._clearAllTransientDetachTimeouts();
-            this._operationInProgress = true;
+            this._startLifecycleOperation();
             this._lifecycle.setStatus({
               status: RoomStatus.Failed,
               error: change.reason,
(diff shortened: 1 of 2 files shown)


sacOO7 (commented):
Can we add a private method to avoid code repetition 
```
  atomic(operation: () => Promise<void>, priority: number): Promise<void> {
      return this._mtx.runExclusive(() => {
        this._operationInProgress = true
        return operation().finally(() =>
          this._operationInProgress = false
        )
      }, priority);
  }
```
So, attach and all other ops would be
```
  attach(): Promise<void> {
    this._logger.trace('RoomLifecycleManager.attach();');
    return this.atomic(async () => {
```

sacOO7:
> > Can we add a private method to avoid code repetition
> 
> We can in the future - though I'd prefer to keep this PR scoped to just the functionality fixes for now.

Sure, I mean, the suggestion is related to the fix. In the future, if we have more atomic operations, we don't need to worry about setting/clearing `_operationInProgress` flag.

sacOO7:
Sure, thanks for the info!

src/core/room-lifecycle-manager.ts
```diff
@@ -404,6 +405,8 @@ export class RoomLifecycleManager {
     try {
       await this._doChannelWindDown(contributor).catch(() => {
         // If in doing the wind down, we've entered failed state, then it's game over anyway
+        // TODO: Another PR, but in the even if we get a failed channel, we still need to do the wind down
```
splindsay-92, line 408: Is the other PR already in review, or still to be started? If it's the former, can you add it to the PR body please?

AndyTWF: Not yet started, but now added an issue link

src/core/room-lifecycle-manager.ts
```diff
@@ -417,27 +420,32 @@ export class RoomLifecycleManager {
       });
     } catch {
       // If an error gets through here, then the room has entered the failed state, we're done.
+      this._endLifecycleOperation();
       return;
     }
 
     // If our problem channel has reattached, then we can retry the attach
-    if (contributor.channel.state === RoomStatus.Attached) {
+    if (contributor.channel.state === 'attached') {
```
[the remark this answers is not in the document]
AndyTWF: They were previously an improper type comparison - the channel states are just strings, not enums (it works as the enum is a superset, but its not a proper comparison)

splindsay-92 (approved):
LGTM

PR #414, opened 2024-11-20
docs: fix `ChatClient` import statement

owenpearson:
the package currently doesn't have a default export


AndyTWF (approved):
LGTM - thanks!

Would you be able to update the website docs too - looks like the same mistake is over there? :)

https://ably.com/docs/chat/setup?lang=javascript

owenpearson:
> Would you be able to update the website docs too - looks like the same mistake is over there? :)

ah good point! have opened a pr now https://github.com/ably/docs/pull/2313

PR #415, opened 2024-11-24
core/room: add channel manager, fix bug where channel options not applied

### LINKED ISSUE #411: Channel options / modes overwritten
At the moment, channel options / modes on shared channels are determined by LWW. So if you have all the features, the options provided by occupancy override those provided by presence.

Shared channels need to be configured such that they merge the modes required by each feature and request the channel only once.

This also requires a new specification point.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/CHA-738) by [Unito](https://www.unito.io)

AndyTWF:
### Context

Closes #411
[CHA-738].

### Description

This change fixes the bug described in #411 whereby channel options are overwritten by successive uses of the same channel, rather than merged together.

In doing this, we also fix the bug in relation to newly introduce spec point CHA-RC3. We were previously depending on the soft-deprecated behaviour in RTS3c which would allow you to update channel options via setOptions.

The change is achieved by adding a channel manager, whose job it is to register all the options for a given channel prior to the first channels.get() call, and call channels.get with the amalgamated options.

### Checklist

* [x] QA'd by the author.

N/A

[CHA-738]: https://ably.atlassian.net/browse/CHA-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/channel-manager.ts
+++ b/src/core/channel-manager.ts
@@ -1,0 +1,46 @@
+import * as Ably from 'ably';
+
+import { Logger } from './logger.js';
+import { DEFAULT_CHANNEL_OPTIONS } from './version.js';
+
+export type ChannelOptionsMerger = (options: Ably.ChannelOptions) => Ably.ChannelOptions;
+
+export class ChannelManager {
+  private readonly _realtime: Ably.Realtime;
+  private readonly _logger: Logger;
+  private readonly _registeredOptions = new Map<string, Ably.ChannelOptions>();
+  private readonly _requestedChannels = new Set<string>();
+
+  constructor(realtime: Ably.Realtime, logger: Logger) {
+    logger.trace('ChannelManager();');
+    this._realtime = realtime;
+    this._logger = logger;
+  }
+
+  mergeOptions(channelName: string, merger: ChannelOptionsMerger): void {
+    this._logger.trace('ChannelManager.registerOptions();', { channelName });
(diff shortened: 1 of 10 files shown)


AndyTWF:
Note, we shouldn't merge this until RAR-815 lands.

src/core/channel-manager.ts
```diff
@@ -0,0 +1,44 @@
...
+  get(channelName: string): Ably.RealtimeChannel {
+    this._logger.trace('ChannelManager.get();', { channelName });
+    this._requestedChannels.add(channelName);
+    return this._realtime.channels.get(
+      channelName,
+      this._registeredOptions.get(channelName) ?? DEFAULT_CHANNEL_OPTIONS,
+    );
+  }
+
+  release(channelName: string): void {
```
owenpearson, line 40: do we need to also `this._requestedChannels.delete(channelName)` here?

owenpearson (commented):
looks good, just one question:

src/core/channel-manager.ts
AndyTWF: Fair shout - done!

owenpearson (approved):
looks good!

JoaoDiasAbly (approved):

PR #420, opened 2024-11-29
core: change direction to orderBy for history, use enum

AndyTWF:
### Description

As already done for Swift/Kotlin, this change renames the direction parameter of the history parameters to orderBy. Instead of using string values, we now use an enum.

The ChatApi now does a mapping to the backend format (which itself will change at a later date).

### Checklist

* [x] QA'd by the author.

N/A

### DIFF
--- a/src/core/chat-api.ts
+++ b/src/core/chat-api.ts
@@ -2,13 +2,14 @@ import * as Ably from 'ably';
 
 import { Logger } from './logger.js';
 import { DefaultMessage, Message, MessageHeaders, MessageMetadata, MessageOperationMetadata } from './message.js';
+import { OrderBy } from './messages.js';
 import { OccupancyEvent } from './occupancy.js';
 import { PaginatedResult } from './query.js';
 
 export interface GetMessagesQueryParams {
   start?: number;
   end?: number;
-  direction?: 'forwards' | 'backwards';
+  orderBy?: OrderBy;
   limit?: number;
   /**
    * Serial indicating the starting point for message retrieval.
@@ -19,6 +20,14 @@ export interface GetMessagesQueryParams {
    */
   fromSerial?: string;
 }
+
+/**
+ * In the REST API, we currently use the `direction` query parameter to specify the order of messages instead
+ * of orderBy. So define this type for conversion purposes.
+ */
+type ApiGetMessagesQueryParams = Omit<GetMessagesQueryParams, 'orderBy'> & {
+  direction?: 'forwards' | 'backwards';
+};
 
 export interface CreateMessageResponse {
   serial: string;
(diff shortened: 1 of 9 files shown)


src/core/messages.ts
```diff
@@ -39,6 +39,21 @@ const MessageActionsToEventsMap: Map<ChatMessageActions, MessageEvents> = new Ma
   [ChatMessageActions.MessageDelete, MessageEvents.Deleted],
 ]);
 
+/**
+ * The order in which results should be returned when performing a paginated query (e.g. message history).
+ */
+export enum ResultOrder {
```
vladvelici, line 45: `OrderBy`?

Nitpick, not strong opinion.

Just thinking it might be easier to remember to type `messages.get({orderBy: OrderBy.NewestFirst})` rather than `messages.get({orderBy: ResultOrder.NewestFirst})`.

Another nitpick: should we define this in `chat-api.ts` instead?

AndyTWF: Worth bringing up with ecosystems (that's where `ResultOrder` originates from) - once the DR is agreed and we definitely do this, we can deffo consider this change :)

AndyTWF: Changed to OrderBy

src/core/chat-api.ts
```diff
@@ -91,9 +100,18 @@ export class ChatApi {
...
   async getMessages(roomId: string, params: GetMessagesQueryParams): Promise<PaginatedResult<Message>> {
     roomId = encodeURIComponent(roomId);
-    return this._makeAuthorizedPaginatedRequest<Message>(`/chat/v2/rooms/${roomId}/messages`, params).then((data) => {
-      return this._recursivePaginateMessages(data);
-    });
+
+    // convert the params into internal format
+    const apiParams: ApiGetMessagesQueryParams = { ...params };
+    if (params.orderBy) {
+      apiParams.direction = params.orderBy === OrderBy.OldestFirst ? 'forwards' : 'backwards';
```
owenpearson, line 107: if someone is using plain javascript without type hints and puts 'forwards' then this will default to 'backwards' since it's only checking whether the param === 'oldestFirst'. i'd recommend using a switch statement and throwing an error if the param isn't a valid `OrderBy` member

src/core/chat-api.ts
```diff
@@ -91,9 +100,18 @@ export class ChatApi {
...
+    }
+
+    return this._makeAuthorizedPaginatedRequest<Message>(`/chat/v2/rooms/${roomId}/messages`, apiParams).then(
```
owenpearson, line 110: this is an async function so can just await this instead of mixing async/await with promise api?

owenpearson (changes requested):

owenpearson (approved):
lgtm!

PR #445, opened 2025-01-07
Message reactions POC

vladvelici:
### Context

- POC for message reactions
- For the POC, this uses the realtime HTTP endpoint instead of a chat-specific one

This allows
- Adding a reaction, and allows for multiple reactions of any type
- Adding a new reaction and at the same time removing all other reactions (to facilitate "react only once" behaviour)
- Scores for each reaction. Scores are added up in summaries.
- Summaries give you total score, total number of reactions, number of users who reacted, and the score and total for each clientId that has reacted.
- Removing all reactions for a message for the current user (not in demo app but supported)
- Removing all reactions for a single emoji for a message for the current user (in demo app as well)

The UI in the demo app shows all options explicitly. In a real application different features will be showed differently. E.g. adding scores might be a through a long-press on the emoji.

image

### Likely will be extracted in different PRs
- Introduction of a Message Window abstraction to simplify keeping a list of messages up to date
- `message.with(event)` method for applying an event to get a new message version

### Does not yet have:
- fetching reactions from history, you only see reactions if you/someone reacts while you're online
- the correct way to subscribe to reaction summaries via `room.messages.reactions.subscribeSummaries` or individual reactions via `room.messages.reactions.subscribe()`. It still goes through `room.messages.subscribe()`.

JIRA:
- https://ably.atlassian.net/browse/CHA-635
- https://ably.atlassian.net/browse/CHA-636
- https://ably.atlassian.net/browse/CHA-619
- https://ably.atlassian.net/browse/CHA-702

### Checklist

* [ ] QA'd by the author.

### Testing Instructions

- React to messages
- MessageWindow is used in demo app so you can see it in action

### DIFF
--- a/src/core/chat-api.ts
+++ b/src/core/chat-api.ts
@@ -141,6 +141,7 @@ export class ChatApi {
         (message.createdAt as Date | undefined) ? new Date(message.createdAt) : new Date(message.timestamp),
         new Date(message.timestamp),
         message.operation,
+        message.reactions,
       );
     };
 
--- a/src/core/chat-api.ts
+++ b/src/core/chat-api.ts
@@ -202,15 +202,40 @@ export class ChatApi {
     );
   }
 
-  async reactToMessage(roomId: string, serial: string, reaction: string): Promise<void> {
+  async addMessageReaction(roomId: string, serial: string, reaction: string, score : number = 1, unique : boolean = false): Promise<void> {
     roomId = encodeURIComponent(roomId);
     return this._makeAuthorizedRequest(`/channels/${roomId}::$chat::$chatMessages/messages`, 'POST', {
-      action: 4,
-      data: reaction,
+      action: 5,
+      data: JSON.stringify({emoji: reaction, score: score, unique: unique}),
       refType: 'reaction:emoji.v1',
       refSerial: serial,
     }).then((response) => {
       console.log('response from sending reaction:', response);
+    });
+  }
+
+  async removeAllMessageReactions(roomId: string, serial: string): Promise<void> {
+    roomId = encodeURIComponent(roomId);
+    return this._makeAuthorizedRequest(`/channels/${roomId}::$chat::$chatMessages/messages`, 'POST', {
+      action: 6,
+      data: "",
+      refType: 'reaction:emoji.v1',
+      refSerial: serial,
+    }).then((response) => {
+      console.log('response from removing all reactions:', response);
+    });
+  }
+
+
+  async removeMessageReaction(roomId: string, serial: string, reaction: string): Promise<void> {
+    roomId = encodeURIComponent(roomId);
+    return this._makeAuthorizedRequest(`/channels/${roomId}::$chat::$chatMessages/messages`, 'POST', {
+      action: 6,
+      data: JSON.stringify({emoji: reaction}),
+      refType: 'reaction:emoji.v1',
+      refSerial: serial,
+    }).then((response) => {
+      console.log(`response from removing '${reaction}' reaction:`, response);
     });
   }
 
--- a/src/core/chat-api.ts
+++ b/src/core/chat-api.ts
@@ -2,7 +2,7 @@ import * as Ably from 'ably';
 
 import { Logger } from './logger.js';
 import { DefaultMessage, Message, MessageHeaders, MessageMetadata, MessageOperationMetadata } from './message.js';
-import { OrderBy } from './messages.js';
+import { MessageReactionMode, OrderBy } from './messages.js';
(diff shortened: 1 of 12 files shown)


src/core/message-events.ts
```diff
@@ -0,0 +1,121 @@
...
+   */
+  type: MessageEvents;
+
+  /**
+   * A shorthand to get the relevant message serial. For message events like
+   * create, update and delete this is the message serial. For annotations
+   * (eg. reactions) this is the `serial` of the relevant chat message, usually
+   * found under `refSerial`.
+   */
+  get messageSerial(): string;
```
splindsay-92, line 16: I'm not so sure about conflating the two here, unless we decide to call it something like `parentSerial` ? I'll have more of a think on this, perhaps we can do some funky union type

vladvelici: I guess the word `message` has too many meanings. If it was an pubsub message, totally confusing. If we talk about a chat message (my intention), it should be clear it's the serial of the chat message (and for reactions it's the serial of the message it refers to).

The purpose is to be able to get the serial of the chat message for any event easily (no type-casting, no switch statements). Pseudocode:
```
subscribeAll(event => {
  storedMessage = findMessageBySerial(event.messageSerial)
  updateMessage(event.messageSerial, storedMessage.apply(event))
});
```

vladvelici: Can definitely rename to `parentSerial`

src/core/messages.ts
```diff
@@ -285,6 +261,11 @@ export interface Messages extends EmitsDiscontinuities {
    */
   update(message: Message, update: UpdateMessageParams, details?: OperationDetails): Promise<Message>;
 
+  /**
+   * Add a reaction to a message.
+   */
+  react(message: Message, reaction: string): Promise<void>;
```
AndyTWF, line 267: I'd suggest `addReaction`, only because eventually we'll want to remove

src/core/messages.ts
```diff
@@ -578,12 +556,35 @@ export class DefaultMessages
...
+    return this._chatApi.reactToMessage(this._roomId, message.serial, reaction);
+  }
+
   /**
    * @inheritdoc Messages
    */
-  subscribe(listener: MessageListener): MessageSubscriptionResponse {
-    this._logger.trace('Messages.subscribe();');
-    super.on([MessageEvents.Created, MessageEvents.Updated, MessageEvents.Deleted], listener);
+  subscribe(listener: MessageListener | MessageListenerObject): MessageSubscriptionResponse {
```
AndyTWF, line 567: Given we're in the 0.x versions, I'd be comfortable with just making the breaking change - rather than having both versions

src/core/chat-api.ts
```diff
@@ -202,6 +202,18 @@ export class ChatApi {
     );
   }
 
+  async reactToMessage(roomId: string, serial: string, reaction: string): Promise<void> {
+    roomId = encodeURIComponent(roomId);
+    return this._makeAuthorizedRequest(`/channels/${roomId}::$chat::$chatMessages/messages`, 'POST', {
```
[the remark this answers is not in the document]
vladvelici: It's temporary, I wanted to get something working end-to-end first

src/core/message-events.ts
vladvelici: I think it makes development easier if you can get the base message serial without any type-casting in the case of `subscribeAll`: you can ignore events early on in your listener, you can find the base message, and you can call message.apply() without type-casting as well.

If this isn't a use case we care about, we can just remove this field. The typed events already have either `serial` or `refSerial` which also matches realtime.

src/core/message.ts
```diff
@@ -288,4 +333,105 @@ export class DefaultMessage implements Message {
...
   }
+
+  apply(event: AnyMessageEvent): Message {
+    if (event.messageSerial !== this.serial) {
+      throw new ErrorInfo('apply(): Cannot apply event to message, serials do not match', 50000, 500);
+    }
+
+    switch (event.type) {
+      case MessageEvents.Created: {
+        // created events shouldn't get here, we'll treat as no-op
```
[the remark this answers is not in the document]
vladvelici: I agree

src/core/messages.ts
vladvelici: True, will change. `addReaction` and `removeReaction` are better than `react` and `removeReaction`.

src/core/messages.ts
vladvelici: Happy to remove the `MessageListener` part

vladvelici:
Closing, replaced by https://github.com/ably/ably-chat-js/pull/473

PR #447, opened 2025-01-07
demo: Make the chat window full height with a min 500px

vladvelici:


splindsay-92 (approved):
Looks good! I had meant to play around with this a bit to get the CSS correct, so thanks for working on this! :D

PR #451, opened 2025-01-14
Update test data with corrected action values

splindsay-92:
Serial and created at fields were added to the test data since ably-js does not infer these from the version field.

### DIFF
--- a/test/core/helpers.test.ts
+++ b/test/core/helpers.test.ts
@@ -12,7 +12,7 @@ const TEST_ENVELOPED_MESSAGE = {
   clientId: 'user1',
   timestamp: 1719948956834,
   encoding: 'json',
-  action: 1,
+  action: 0,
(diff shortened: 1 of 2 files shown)


test/core/helpers.test.ts
```diff
@@ -9,10 +9,11 @@ import { DefaultReaction } from '../../src/core/reaction.js';
 
 const TEST_ENVELOPED_MESSAGE = {
   id: 'chat:6TP2sA:some-room:219f7afc614af7b:0',
+  serial: '01719948956834-000@108TeGZDQBderu97202638',
```
AndyTWF, line 12: We should probably omit the `serial` here per the original test - it was because of the incorrect `action` that it wasn't parsing it

splindsay-92: Done

AndyTWF (approved):

PR #452, opened 2025-01-14
Bugfix: Fix race condition in usePresence test

splindsay-92:
### Context

* [CHA-782]
* We found a race condition caused by umounting the component in the `usePresence` test. During unmount, we attempt to `leave` presence. Due to how the test is written, unmounting also detaches us from the channel. In the race between the former and the latter, the latter wins and so the `leave` call fails since we are no longer in an `attached` state.

### Description

* Update the test so the Providers do not umount, this means we can test the hooks unmount process without triggering the detach.

### Checklist

* [x] QA'd by the author.

[CHA-782]: https://ably.atlassian.net/browse/CHA-782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/react/hooks/use-presence.integration.test.tsx
+++ b/test/react/hooks/use-presence.integration.test.tsx
@@ -1,6 +1,6 @@
 import { PresenceData, PresenceEvent, PresenceEvents, RoomOptionsDefaults } from '@ably/chat';
 import { cleanup, render, waitFor } from '@testing-library/react';
-import { useEffect } from 'react';
+import React, { useEffect } from 'react';
 import { afterEach, describe, expect, it } from 'vitest';
 
 import { usePresence } from '../../../src/react/hooks/use-presence.ts';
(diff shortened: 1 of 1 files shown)


test/react/hooks/use-presence.integration.test.tsx
```diff
@@ -112,13 +114,16 @@ describe('usePresence', () => {
       presenceEventsRoomTwo,
     );
 
-    // unmount the component
-    unmount();
+    // Remove TestComponent while keeping Providers intact
+    rerender(<Providers>{null}</Providers>);
```
owenpearson, line 118: I think if you do the above suggestion you won't have to stick `null` in here either

owenpearson (approved):
LGTM, feel free to ignore this tiny suggestion:

test/react/hooks/use-presence.integration.test.tsx
splindsay-92: I tried with the above, but the `null` is still needed. Perhaps I am missing something?

PR #457, opened 2025-01-23
Add message.with(event) method for applying message events to a Message

vladvelici:
### Context

[CHA-702]

### Description

Add a new method `with()` on `Message`. `message.with(event)` is a convenience method that produces a new Message with the event applied to it. It can be used to easily apply update and delete events to a message.

Usage example

```typescript
let messages : Message[] = []; // my messages

// when receiving an update or delete event:
const idx = message.findIndex(m => m.serial == event.message.serial);
if (idx !== -1) {
  messages[idx] = messages[idx].with(event);
}
```

When message reactions arrive, this method will also correctly apply reactions events to a message.

---

Another change is that the demo app now shows deleted messages with an "Edit" button. Since we have always supported what I'll call "restore by edit" it makes sense to show it in the UI, as well as to correctly support showing the messages that have been updated after a delete (previously those update events were ignored, considered out of scope for the chat window).

### Checklist

* [x] QA'd by the author.

Read the code including demo app for how it's used.

[CHA-702]: https://ably.atlassian.net/browse/CHA-702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/message.ts
+++ b/src/core/message.ts
@@ -1,6 +1,6 @@
 import { ErrorInfo } from 'ably';
 
-import { ChatMessageActions } from './events.js';
+import { ChatMessageActions, MessageEventPayload, MessageEvents } from './events.js';
 import { Headers } from './headers.js';
 import { Metadata } from './metadata.js';
 import { OperationMetadata } from './operation-metadata.js';
@@ -201,6 +201,17 @@ export interface Message {
    * @throws {@link ErrorInfo} if serials of either message is invalid.
    */
   equal(message: Message): boolean;
+
+  /**
+   * Creates a new message instance with the event applied.
+   *
+   * @param event The event to be applied to the returned message.
+   * @throws {@link ErrorInfo} if the event is for a different message.
+   * @throws {@link ErrorInfo} if the event is a {@link MessageEvents.Created}.
+   * @returns A new message instance with the event applied. If the event is a no-op, such
+   *    as an event for an old version, the same message is returned (not a copy).
+   */
+  with(event: MessageEventPayload): Message;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


src/core/message.ts
```diff
@@ -288,4 +298,20 @@ export class DefaultMessage implements Message {
   equal(message: Message): boolean {
     return this.serial === message.serial;
   }
+
+  with(event: MessageEventPayload): Message {
+    if (event.type === MessageEvents.Created) {
+      throw new ErrorInfo('Cannot apply a created event to a message', 40000, 400);
```
AndyTWF, line 304: We use lowercase for our messages elsewhere, so should follow the same pattern here

src/core/message.ts
```diff
@@ -288,4 +298,20 @@ export class DefaultMessage implements Message {
...
+
+    if (event.message.serial !== this.serial) {
+      throw new ErrorInfo('Cannot apply event for a different message', 40000, 400);
+    }
+
+    if (this.version >= event.message.version) {
+      return this;
+    }
+
+    return event.message;
```
AndyTWF, line 315:
```suggestion
    return this.version >= event.message.version ? this : event.message
```

Ternary perhaps a bit cleaner?

src/core/message.ts
AndyTWF: Also - do we just want to make passing in `MessageEvents.Created` a no-op, rather than an error?

vladvelici: That was my initial attempt https://github.com/ably/ably-chat-js/pull/445#discussion_r1907068897

I think error because if users end up calling with(event) for creates, it suggests they haven't processed the create separately. The error will give a clear signal something's wrong (better DX when compared to new messages not showing and having no errors).

src/core/message.ts
vladvelici: Done

AndyTWF (approved):

splindsay-92 (approved):
LGTM

PR #458, opened 2025-01-27
Poc: single channel presence

splindsay-92:
### Context

* 
### Description

* Describe the changes introduced in this PR
* Provide relevant context or considerations for reviewers.
    * Such as why the changes are necessary,
      or how they affect the system.

### Checklist

* [ ] QA'd by the author.


src/core/presence-data-manager.ts
```diff
@@ -0,0 +1,566 @@
...
+ * It ensures
+ * that multiple callers for presence state retrieval all resolve to the same state if made concurrently while
+ * the state is being fetched.
+ */
+export class DefaultPresenceSetManager {
+  private _queue: QueueElement[] = [];
+  private _isFetching = false;
+  private _retryCount = 0;
+  private _channel: Ably.RealtimeChannel;
+  private _logger: Logger;
```
AndyTWF, line 367:
```suggestion
  private readonly _queue: QueueElement[] = [];
  private _isFetching = false;
  private _retryCount = 0;
  private readonly _channel: Ably.RealtimeChannel;
  private readonly _logger: Logger;
```

src/core/presence-data-manager.ts
```diff
@@ -0,0 +1,566 @@
...
+      return false;
+    }
+    const dataObject = data as {
+      typing?: { isTyping: unknown },
+      presence?: { nonce?: unknown },
+      userCustomData?: unknown
+    }
+
+    // Ensure typing contains a boolean value if present
+    if (dataObject.typing !== undefined && typeof dataObject.typing.isTyping !== 'boolean') {
```
AndyTWF, line 482:
```suggestion
    if (typeof dataObject.typing?.isTyping !== 'boolean') {
```

src/core/presence-data-manager.ts
```diff
@@ -0,0 +1,566 @@
...
+      return false;
+    }
+    // Ensure presence contains a string nonce if present
+    if (dataObject.presence !== undefined && dataObject.presence.nonce !== undefined && typeof dataObject.presence.nonce !== 'string') {
```
AndyTWF, line 486:
```suggestion
    if (typeof dataObject.presence?.nonce !== 'string') {
```

src/core/presence.ts
```diff
@@ -55,6 +81,11 @@ export interface PresenceEvent {
    */
   clientId: string;
 
+  /**
+   * The connectionId of the presence member.
+   */
+  connectionId: string;
```
AndyTWF, line 87: Should we provide some method here to help users with determining unique members in the presence set, e.g. `memberKey()`?

src/core/presence.ts
```diff
@@ -176,6 +223,12 @@ export interface Presence extends EmitsDiscontinuities {
    */
   unsubscribeAll(): void;
 
+  /**
+   * Subscribe to presence set change events, which are emitted when a presence event is received.
+   * @param listener
+   */
+  onPresenceSetChange(listener: (presenceSetChangeEvent: PresenceSetChangeEvent) => void): { off: () => void };
```
AndyTWF, line 230: We should define the return type as some explicitly named type

src/core/presence.ts
```diff
@@ -192,23 +245,73 @@ export class DefaultPresence
...
   private readonly _discontinuityEmitter: DiscontinuityEmitter = newDiscontinuityEmitter();
+  private readonly _onPresenceSetChangeEmitter = new EventEmitter<PresenceSetChangeEventMap>();
+  private readonly _presenceDataContribution: PresenceDataContribution;
+  private readonly _presenceManager: PresenceManager;
 
   /**
    * Constructs a new `DefaultPresence` instance.
    * @param roomId The unique identifier of the room.
    * @param channelManager The channel manager to use for creating the presence channel.
+   * @param presenceManager
```
AndyTWF, line 261: Description

src/core/presence.ts
```diff
@@ -192,23 +245,73 @@ export class DefaultPresence
...
+    this._listenForDiscontinuities();
+  }
+
+  /**
+   * Listens for discontinuities and clears the presence members when a discontinuity is detected.
+   * @private
+   */
+  private _listenForDiscontinuities() {
+    this.onDiscontinuity(() => {
+      this._logger.warn('Presence._listenForDiscontinuities(); Discontinuity detected, clearing presence members');
```
AndyTWF, line 290: I think this can just be a debug log

src/core/presence.ts
```diff
@@ -265,10 +366,11 @@ export class DefaultPresence
    */
   async enter(data?: PresenceData): Promise<void> {
```
AndyTWF, line 367: Are `enter` and `leave` still the right method names, seeing as they no longer directly correspond to the underlying channel events?

AndyTWF: Should we instead be doing `setUserData`, etc?

src/core/presence.ts
```diff
@@ -350,31 +465,242 @@ export class DefaultPresence
...
+   */
+  private _setPresenceMembersFromMessages(messages: PresenceMessage[]): void {
+    this._logger.trace('Presence._setPresenceMembersFromMessages();', { messages });
+    // Clear the current presence members
+    this._presenceMembers.clear();
+    for (const message of messages) {
+      this._setPresenceMember(message);
+    }
+    this._logger.trace('Presence._setPresenceMembersFromMessages(); New presence members set', {
+      members: [...this._presenceMembers.values()],
```
AndyTWF, line 502: Do we need to log all of this? Perhaps just the length? Quite expensive to log on every change

src/core/presence.ts
```diff
@@ -350,31 +465,242 @@ export class DefaultPresence
...
+    }
+  }
+
+  /**
+   * Generates a unique key for a presence member based on the clientId and connectionId.
+   * @param clientId
+   * @param connectionId
+   * @private
+   */
+  private _keyFromClientAndConnectionId(clientId: string, connectionId: string): string {
```
AndyTWF, line 535: Make this function take an interface with these properties and just have a single argument? Makes shorthanding the code easier.

src/core/presence.ts
```diff
@@ -350,31 +465,242 @@ export class DefaultPresence
...
+          action: PresenceEvents.Update,
+          clientId: current.clientId,
+          connectionId: current.connectionId,
+          timestamp: current.timestamp,
+          data: chatPresenceData.userCustomData,
+        };
+        break;
+      }
+      case PresenceEvents.Enter: {
+        this._logger.debug('Presence._handlePresentMember(); Member was already presence', { current });
```
AndyTWF, line 623:
```suggestion
        this._logger.debug('Presence._handlePresentMember(); Member was already present', { current });
```

src/core/room-options.ts
```diff
@@ -1,33 +1,25 @@
...
 export const DefaultRoomOptions = {
   /**
    * The default presence options for a chat room.
+   * Includes default options for entering/subscribing to presence and typing timeout.
    */
   presence: {
-    /**
-     * The client should be able to enter presence.
-     */
+    typingOptions: TypingOptionsDefaults,
```
AndyTWF, line 19: Would just call this `typing`

src/core/room.ts
```diff
@@ -178,13 +183,26 @@ export class DefaultRoom implements Room {
...
+        logger,
+      );
       features.push(this._presence);
     }
 
-    if (options.typing) {
+    if (options.presence?.typingOptions) {
       this._logger.debug('enabling typing on room', { roomId });
-      this._typing = new DefaultTyping(roomId, options.typing, channelManager, realtime.auth.clientId, logger);
+      this._typing = new DefaultTyping(
```
AndyTWF, line 198: If we're going to conceptualise this as being a part of presence, then it should be structured as `room.presence.typing`, or just having methods directly on `room.presence`

src/core/typing.ts
```diff
@@ -117,59 +134,190 @@ export class DefaultTyping
...
    * @inheritDoc
    */
-  get(): Promise<Set<string>> {
+  async get(): Promise<Set<string>> {
     this._logger.trace(`DefaultTyping.get();`);
-    return this._channel.presence.get().then((members) => new Set<string>(members.map((m) => m.clientId)));
+    const members = await this._presenceManager.getPresenceSet()
+    return new Set<string>(
+      members
+        .filter((m) => (m.data ? (m.data as ChatPresenceData).typing?.isTyping : false))
```
AndyTWF, line 318: `m` and `m_1` aren't descriptive. Can we call it something like `member`?

demo/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "dependencies": {
     "@ably/chat": "file:..",
-    "ably": "^2.6.3",
```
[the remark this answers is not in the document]
vladvelici: FYI in case you try and hit the same issue as me. After using a branch and then updating it, npm refused to re-fetch. Tried the `rm -rf`, tried `npm cache clean` and used the `--force` too... Eventually, using commit hash directly worked as a workaround.

src/core/presence.ts
splindsay-92: So I think I see merits for both, enter/update/leave before was controlling realtime presence, now they control just the chat presence portion of the underlying presence payload. Doing so keeps the API very similar to ably-js presence, but has the downside of potentially confusing what you see in integrations (enter in chat might correspond to an update on the channels presence)

The alternative in events for user data would be something like this I suppose..
`userdata.set`, `userdata.update` and `userdata.remove` which really is just presence events renamed. The presence payload we send would likely look very similar to what we have, just instead of `presence?: {}` we would have `userdata?:{}`. I think the change to `userdata`  highlights that presence now contains multiple features, but beyond that, there isn't a big difference imo.

AndyTWF:
I'm closing this, as the POC is no longer relevant.

PR #468, opened 2025-02-05
make npm lint:fix to also format

vladvelici:
I think after we run lint:fix the expectation is that the code is in acceptable shape (or command fails/shows some errors), so running format at the end of linting.


splindsay-92 (approved):
LGTM

PR #473, opened 2025-02-11
Message reactions

vladvelici:
### Context

- https://ably.atlassian.net/browse/CHA-635
- https://ably.atlassian.net/browse/CHA-636
- https://ably.atlassian.net/browse/CHA-619

### Description

There are still things missing, this is a work in progress.

This PR replaces the earlier reactions prototype https://github.com/ably/ably-chat-js/pull/445 (now closed). The previous prototype was using a single refType with a more complex reaction payload to determine the behaviour of each reaction.

The current approach uses three different refTypes, one for each desired behaviour:
1. one reaction per message per user (in total)
2. at most one reaction per message per user for each emoji
3. any reactions (and single ones can count more than 1 via score)

This has the advantage of having separate summary counts for each refType, and also knowing that everyone who participates with reactions in a given refType is subject to the same rules.

Previously (and what many alternative Chat SDKs do) the client was choosing how to react. The application could offer a default (eg via UI) but someone using the dev console (or another client implementation) could add reactions in any mode they liked and have them counted up incorrectly / unfairly to everyone else. Now if someone attempts to use a different refType (eg to react many times in a one-reaction-per-message setting), their reactions might work in the separate refType but nobody else will see them since they'll be watching another refType altogether. Capabilities might help restrict this further in the future.

### Left to do

- [x] Fix tests
- [x] Write tests for new things
- [x] Individual annotation events
- [x] Configuration for whether to receive individual annotations for reactions or not (whether chat client sets the annotation_subscribe channel mode)
- [x] Check/write/update typedocs
- [x] Remove annotation deletes (will come in future PR)

### Checklist

* [x] QA'd by the author.


src/core/messages.ts
```diff
@@ -277,6 +293,130 @@ export interface Messages extends EmitsDiscontinuities {
...
+  }
+
+  private _processMessageEvent(event: Ably.InboundMessage) {
+    // only process summary events
+    if (event.action !== ChatMessageActions.MessageAnnotationSummary) {
+      return;
+    }
+    // they must have a refSerial
+    if (!event.refSerial) {
+      return;
```
AndyTWF, line 346: Worth a `warn` log here, as this is unintended behaviour?

src/core/messages.ts
```diff
@@ -277,6 +293,130 @@ export interface Messages extends EmitsDiscontinuities {
...
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/prefer-optional-chain, @typescript-eslint/no-unsafe-assignment
+    const single: Record<string, SingleReactionSummary> = (event.summary || {})[ReactionRefType.Single] || {};
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/prefer-optional-chain, @typescript-eslint/no-unsafe-assignment
+    const distinct: Record<string, DistinctReactionSummary> = (event.summary || {})[ReactionRefType.Distinct] || {};
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/prefer-optional-chain, @typescript-eslint/no-unsafe-assignment
+    const counter: Record<string, CounterReactionSummary> = (event.summary || {})[ReactionRefType.Counter] || {};
+
+    this._emitter.emit('summary', {
+      type: 'message-reaction-summary',
```
AndyTWF, line 361: Should we make this an enum? Its less error prone for users

src/core/messages.ts
```diff
@@ -277,6 +293,130 @@ export interface Messages extends EmitsDiscontinuities {
...
+    // they must have a refSerial
+    if (!event.refSerial) {
+      return;
+    }
+    // they must have a version
+    if (!event.version) {
+      return;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/prefer-optional-chain, @typescript-eslint/no-unsafe-assignment
```
AndyTWF, line 353: We should make this typescript-happy

src/core/messages.ts
```diff
@@ -277,6 +293,130 @@ export interface Messages extends EmitsDiscontinuities {
...
+      timestamp: new Date(event.timestamp),
+      refSerial: event.refSerial,
+      version: event.version,
+      single: single,
+      distinct: distinct,
+      counter: counter,
+    });
+  }
+
+  add(message: { serial: string }, refType: ReactionRefType, reaction: string, count?: number): Promise<void> {
```
AndyTWF, line 371: I think that having the refType on a per-reaction basis is probably overly-configurable. I can't think of any Chat apps that mix two schemes, and I also don't think a scheme change will come up regularly.

I would suggest that the message reaction mode is controlled by `RoomOptions` and set once at initialisation. If we ever get a case where it needs to be per-message, we can add an optional parameter later.

src/core/messages.ts
```diff
@@ -277,6 +293,130 @@ export interface Messages extends EmitsDiscontinuities {
...
+  }
+
+  /**
+   * Subscribe to message reaction summaries. Use this to keep message reaction
+   * counts up to date efficiently in the UI.
+   * @param listener
+   * @returns
+   */
+  subscribe(listener: MessageReactionListener): Unsubscribable {
+    const unique = (event: MessageReactionSummaryEvent) => {
```
AndyTWF, line 390: We don't do this elsewhere - so we should pick an approach globally and use it

src/core/messages.ts
```diff
@@ -305,23 +448,44 @@ export class DefaultMessages
    * @param clientId The client ID of the user.
    * @param logger An instance of the Logger.
    */
-  constructor(roomId: string, channelManager: ChannelManager, chatApi: ChatApi, clientId: string, logger: Logger) {
+  constructor(
+    roomId: string,
+    options: MessageOptions | undefined,
```
AndyTWF, line 453: Should we make this always defined and set a default?

src/core/room-options.ts
```diff
@@ -38,8 +38,32 @@ export const DefaultRoomOptions = {
...
+   * Whether to enable receiving raw individual message reactions from the
+   * realtime channel. The default is false, meaning that individual message
+   * reactions are not received. Set to true if subscribing to raw message
+   * reactions.
+   *
+   * Note reaction summaries (aggregates) are always available regardless of
+   * this setting.
+   * @defaultValue false
+   */
+  RawMessageReactions?: boolean;
```
AndyTWF, line 64:
```suggestion
  rawMessageReactions?: boolean;
```

src/core/room.ts
```diff
@@ -239,6 +247,15 @@ export class DefaultRoom implements Room {
       manager.mergeOptions(DefaultPresence.channelName(this._roomId), DefaultPresence.channelOptionMerger(options));
     }
 
+    manager.mergeOptions(messagesChannelName(this._roomId), (opts) => {
+      if (opts.modes) {
+        opts.modes.push('ANNOTATION_PUBLISH');
+      } else {
+        opts.modes = ['publish', 'subscribe', 'presence_subscribe', 'presence', 'ANNOTATION_PUBLISH'];
```
AndyTWF, line 254: Is the difference in casing intentional here?

src/react/hooks/use-messages.ts
```diff
@@ -185,13 +223,57 @@ export const useMessages = (params?: UseMessagesParams): UseMessagesResponse =>
     ).unmount();
   }, [context, logger, onDiscontinuityRef]);
 
+  useEffect(() => {
+    if (!params?.reactionsListener) return;
+    return wrapRoomPromise(
+      context.room,
+      (room) => {
+        if (!params.reactionsListener) {
```
AndyTWF, line 231: Is this needed? This is checked above.

src/core/room.ts
vladvelici: I think we have some issue with this, actually.

In tests, the casing is always uppercase, so I'm assuming this never runs (only the first condition).

Previously I put it lowercase everywhere, and in test the values look all uppercase except the new one I set.

Maybe change them all to upper? Or I can dig some more to see where these come from?

src/core/room-options.ts
vladvelici: done

src/core/messages.ts
vladvelici: Done.

src/core/message.ts
```diff
@@ -232,8 +251,21 @@ export class DefaultMessage implements Message {
     public readonly createdAt: Date,
     public readonly timestamp: Date,
     public readonly operation?: Operation,
+    reactions?: MessageReactions,
```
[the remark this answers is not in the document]
vladvelici: Done

src/react/hooks/use-messages.ts
vladvelici: I thought yes, because params might change before promise resolve. I saw that this isn't the case since the wrapper doesn't resolve anymore in that case.

But still, the check pleases the linter without a typecast so I'd leave it it.

src/core/chat-api.ts
```diff
@@ -202,7 +209,24 @@ export class ChatApi {
...
   }
 
-  async getOccupancy(roomId: string): Promise<OccupancyEvent> {
+  addMessageReaction(roomId: string, serial: string, data: MessageReactionData): Promise<void> {
+    const encodedSerial = encodeURIComponent(serial);
+    roomId = encodeURIComponent(roomId);
+    return this._makeAuthorizedRequest(`/chat/v2/rooms/${roomId}/messages/${encodedSerial}/reactions`, 'POST', data);
+  }
+
+  deleteMessageReaction(roomId: string, serial: string, data: Omit<MessageReactionData, 'count'>): Promise<void> {
```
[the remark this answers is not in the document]
vladvelici: I'll remove everything except the events related to delete, and add them back in another PR so we're not blocked with this one.

src/core/messages.ts
```diff
@@ -277,6 +293,130 @@ export interface Messages extends EmitsDiscontinuities {
...
+}
+
+export type MessageReactionListener = (event: MessageReactionSummaryEvent) => void;
+export type MessageRawReactionListener = (event: MessageReactionEvent) => void;
+
+export interface Unsubscribable {
+  unsubscribe: () => void;
+}
+
+interface Reactions {
```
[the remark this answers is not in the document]
vladvelici: Done

src/core/messages.ts
```diff
@@ -277,6 +293,130 @@ export interface Messages extends EmitsDiscontinuities {
...
+  refType: string;
+  reaction: string;
+  clientId: string;
+  timestamp: Date;
+}
+
+export type MessageReactionListener = (event: MessageReactionSummaryEvent) => void;
+export type MessageRawReactionListener = (event: MessageReactionEvent) => void;
+
+export interface Unsubscribable {
```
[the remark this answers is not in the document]
vladvelici: Done

src/core/messages.ts
vladvelici: Done

src/core/messages.ts
vladvelici: done

src/core/chat-api.ts
```diff
@@ -85,6 +92,12 @@ interface DeleteMessageParams {
   metadata?: MessageOperationMetadata;
 }
 
+export interface AddMessageReactionParams {
```
AndyTWF, line 95: Appreciate its internal, but might be worth a comment on the type?

src/core/messages.ts
```diff
@@ -260,6 +285,215 @@ export interface Messages extends EmitsDiscontinuities {
...
+export interface MessagesReactions {
+  /**
+   * Add a message reactions
+   * @param message The message to react to.
+   * @param type The type of reaction reference.
+   * @param reaction The reaction to add.
+   * @param count The count of the reaction for types that support it, default 1.
+   * @returns A promise that resolves when the reaction is added.
+   */
+  add(message: { serial: string }, type: MessageReactionType, reaction: string, count?: number): Promise<void>;
```
AndyTWF, line 313: Should we make `type` part of the room options? I can't think realistically of many instances whereby anyone will switching adhoc or using two at once. Even if we keep the in-method version as an optional parameter on the end, I think having the default in the options is valuable.

AndyTWF: Generally speaking - as we've got quite a few arguments here, with a few optionals, might be worth making a `MessageReactionParams` type that wraps them all?

src/core/messages.ts
```diff
@@ -260,6 +285,215 @@ export interface Messages extends EmitsDiscontinuities {
...
+    private readonly _channel: Ably.RealtimeChannel,
+  ) {
+    void _channel.subscribe(this._processMessageEvent.bind(this));
+    if (this._options?.rawMessageReactions) {
+      void _channel.annotations.subscribe(this._processAnnotationEvent.bind(this));
+    }
+  }
+
+  private _processAnnotationEvent(event: Ably.Annotation) {
+    if (!event.refSerial) {
```
AndyTWF, line 357: Should we be logging this as a `warn`?

src/core/messages.ts
```diff
@@ -288,23 +525,44 @@ export class DefaultMessages
...
   }
 
   /**
    * Creates the realtime channel for messages.
    */
   private _makeChannel(roomId: string, channelManager: ChannelManager): Ably.RealtimeChannel {
-    const channel = channelManager.get(messagesChannelName(roomId));
+    const channelName = messagesChannelName(roomId);
+
+    if (this._options.rawMessageReactions) {
```
AndyTWF, line 556: At this point it may be too late to do the merge, as `channels.get` may have already been called - we usually do this in `room.getChannelManager`

src/core/room.ts
AndyTWF: Lets dig more to see where these come from - though I believe that if we look at the underlying ably-js types, they should be lowercase?

src/core/messages.ts
```diff
@@ -260,6 +285,215 @@ export interface Messages extends EmitsDiscontinuities {
...
+    }
+
+    let reaction = event.data as string;
+    let count: number | undefined;
+    if (reactionType === MessageReactionType.Multiple) {
+      const data = reaction as unknown as { count?: number; reaction: string };
+      reaction = data.reaction;
+      if (eventType === MessageReactionEvents.Create) {
+        count = data.count;
+        if (!count || count < 1) {
```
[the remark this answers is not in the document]
vladvelici: Wanted to catch negatives too

src/core/chat-api.ts
vladvelici: done

src/core/messages.ts
vladvelici: added debug logs. I don't think they should be `warn` since it's nothing wrong that this user did, they just received something that's not complete, so will be ignored.

Seems we do _debug_ in DefaultMessages `_processEvent` as well.

src/core/messages.ts
AndyTWF: > asymmetric since you'll have a default for sending but not for receiving. Or do you suggest we add a default for receiving as well somehow?

On the receiving end, we give everyone everything, right? I don't think that sort of asymmetry is too bad. I'm just wary if we're requiring someone to pass in a type every single time its just something else they have to do wherein 99% of use-cases, one type will be all they need.

> Since there's only one optional, count, I'd argue to keep it as-is. If we make type optional as well then I'd change it to add(message, reaction, { type?, count? })

Just thinking of what we do with `Messages` - having an object-type param now would set us up for a similar API in the future

vladvelici: We had a call about this, decision is:

1. Add the a default `type` as room option, and make `type` parameter optional in `reactions.add()`.
2. Use an object as parameter to match what we do with `room.messages.send()` and other methods.
3. On the receiving end (so `Message` interface) so make no change: you still have `message.reactions.<type>[reaction]`

We said no to:
1. Adding 3 (so 6 with deletes included) add reaction methods, one set for each type to avoid configs/defaults.
2. Adding a default `message.reactions` (to have the default type) along with `message.reactionTypes` (like current one with access to all), because typing the `message.reactions` correctly requires to use either `any`, `unknown` or generics at least up to the Room level but potentially to ChatClient.

src/core/messages-reactions.ts
```diff
@@ -0,0 +1,316 @@
...
+    void _channel.subscribe(this._processMessageEvent.bind(this));
+    if (this._options?.rawMessageReactions) {
+      void _channel.annotations.subscribe(this._processAnnotationEvent.bind(this));
+    }
+    this._defaultType = this._options?.defaultMessageReactionType ?? MessageReactionType.Distinct;
+  }
+
+  private _processAnnotationEvent(event: Ably.Annotation) {
+    if (!event.messageSerial) {
+      this._logger.debug(
```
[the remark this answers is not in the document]
vladvelici: I didn't think of that before, changed to `warn`.

src/core/events.ts
```diff
@@ -99,3 +96,152 @@ export interface MessageEvent {
...
+   */
+  Summary = 'reaction.summary',
+}
+
+/**
+ * Represents an individual message reaction event.
+ */
+export interface MessageReactionRawEvent {
+  /** Whether reaction was added or removed */
+  type: MessageReactionEvents.Create | MessageReactionEvents.Delete;
```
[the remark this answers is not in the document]
vladvelici: changed

AndyTWF (approved):

src/core/room-options.ts
```diff
@@ -39,8 +41,42 @@ export const AllFeaturesEnabled = {
...
+ */
+export interface MessageOptions {
+  /**
+   * Whether to enable receiving raw individual message reactions from the
+   * realtime channel. Set to true if subscribing to raw message reactions.
+   *
+   * Note reaction summaries (aggregates) are always available regardless of
+   * this setting.
+   *
+   * @defaultValue true
```
maratal, line 65: Shouldn't this be `false`?

PR #479, opened 2025-02-28
refactor: update message version comparison methods

AndyTWF:
### Description

- Modify message version comparison methods to return boolean instead of throwing errors.
- Add new methods like `isSameAs` and rename existing methods for clarity.

### Checklist

* [x] QA'd by the author.

N/A


src/core/message.ts
```diff
@@ -261,28 +264,25 @@ export class DefaultMessage implements Message {
     return this.isDeleted ? this.timestamp : undefined;
   }
 
-  versionBefore(message: Message): boolean {
-    // Check to ensure the messages are the same before comparing operation order
+  isOlderVersionOf(message: Message): boolean {
```
vladvelici, line 267: I think having both `isOlderVersionOf` and `isNewerVersionOf` might create minor confusion if negating the comparison

`c1 = a.isNewerVersionOf(b)` is not the same as `c2 = !a.isOlderVersionOf(b)` because if `a` and `b` are different messages, `c1` will be false but `c2` will be true.

Maybe add a `@remark` not to negate those? Or maybe it's a non-issue... The docs do explain what the functions do.

AndyTWF: I think a remark to say don't negate them sounds good.

AndyTWF: Remark added

src/core/message.ts
```diff
@@ -148,28 +148,40 @@ export interface Message {
...
+   * **Note** that negating this function does not mean that the message is a newer
+   * version of the same message, as the two may be different messages entirely.
+   *
+   * ```ts
+   *  !message.isOlderVersionOf(other) !== message.isNewerVersionOf(other)
+   * ```
    * @param message The message to compare against.
-   * @returns true if the version of this message is before the given message.
-   * @throws {@link ErrorInfo} if both message serials do not match.
+   * @returns false if the messages are different or if this version is not older.
```
vladvelici, line 160:
```suggestion
   * @returns true if this version is older and the messages are the same
```

Document the `true` condition?

src/core/message.ts
```diff
@@ -148,28 +148,40 @@ export interface Message {
...
+   * version of the same message, as the two may be different messages entirely.
+   *
+   * ```ts
+   *  !message.isNewerVersionOf(other) !== message.isOlderVersionOf(other)
+   * ```
+   *
    * @param message The message to compare against.
-   * @returns true if the version of this message is after the given message.
-   * @throws {@link ErrorInfo} if both message serials do not match.
+   * @returns false if the messages are different or if this version is not newer.
```
vladvelici, line 175: Same as above?

vladvelici (approved):
Left minor comments, approving now.

src/core/message.ts
AndyTWF: done

src/core/message.ts
AndyTWF: done

ttypic (approved):
Looks great!

PR #480, opened 2025-02-28
Refactor: Typing with heartbeats

splindsay-92:
### Context

* As part of the move to a single channel, we needed to rethink how typing works, since both typing and chat presence would have to use the same underlying realtimePresence instance. We played with the idea of a shared presence set, but this added a bit too much complexity in the SDK. 
* Instead, we decided to use heartbeats that will be published similar to the way room reactions are currently. 
* More information can be found in [CHADR-093]

### Description

* Updated the typing class to remove presence and use channel.publish() to send heartbeats

### Checklist

* [x] QA'd by the author.

* Open the demo app and play around with the typing component.

### DIFF
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -76,18 +76,18 @@ export enum TypingEvents {
   /**
    * Event triggered when a user is typing.
    */
-  Start = 'typing.start',
+  Start = 'typing.started',
 
   /**
    * Event triggered when a user stops typing.
    */
-  Stop = 'typing.stop',
+  Stop = 'typing.stopped',
 }
 
 /**
  * Represents a typing event payload.
  */
-export interface TypingEventPayload {
+export interface TypingEvent {
   /**
    * Get a set of clientIds that are currently typing.
    */
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -69,9 +69,39 @@ export enum PresenceEvents {
   Present = 'present',
 }
 
+/**
+ * All typing events.
+ */
 export enum TypingEvents {
-  /** The set of currently typing users has changed. */
-  Changed = 'typing.changed',
+  /**
+   * Event triggered when a user is typing.
+   */
+  Start = 'typing.start',
+
+  /**
+   * Event triggered when a user stops typing.
+   */
+  Stop = 'typing.stop',
+}
+
+/**
+ * Represents a typing event payload.
+ */
+export interface TypingEventPayload {
+  /**
+   * Get a set of clientIds that are currently typing.
+   */
+  get currentlyTyping(): Set<string>;
+
+  /**
+   * Get the client ID of the user who stopped/started typing.
+   */
+  clientId: string;
+
+  /**
+   * Type of the event.
+   */
+  type: TypingEvents;
 }
 
 /**
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -85,7 +85,7 @@ export enum TypingEvents {
 }
 
 /**
- * Represents a typing event payload.
+ * Represents a change in the state of current typers.
  */
 export interface TypingEvent {
   /**
@@ -94,7 +94,7 @@ export interface TypingEvent {
   currentlyTyping: Set<string>;
 
   /**
-   * Represents the change that resulted in the typing event.
+   * Represents the change that resulted in the new set of typers.
    */
   change: {
     /**
@@ -103,7 +103,7 @@ export interface TypingEvent {
     clientId: string;
 
     /**
-     * Type of the event.
+     * Type of the change. Either `typing.started` or `typing.stopped`.
      */
     type: TypingEvents;
   };
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -89,16 +89,16 @@ export enum TypingEvents {
  */
 export interface TypingEvent {
   /**
-   * Get a set of clientIds that are currently typing.
+   * The set of clientIds that are currently typing.
    */
-  get currentlyTyping(): Set<string>;
+  currentlyTyping: Set<string>;
 
   /**
    * Represents the change that resulted in the typing event.
    */
   change: {
     /**
-     * Get the client ID of the user who stopped/started typing.
+     * The client ID of the user who stopped/started typing.
      */
     clientId: string;
 
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -94,14 +94,19 @@ export interface TypingEvent {
   get currentlyTyping(): Set<string>;
 
   /**
-   * Get the client ID of the user who stopped/started typing.
+   * Represents the change that resulted in the typing event.
    */
-  clientId: string;
+  change: {
+    /**
+     * Get the client ID of the user who stopped/started typing.
+     */
+    clientId: string;
 
-  /**
-   * Type of the event.
-   */
-  type: TypingEvents;
+    /**
+     * Type of the event.
+     */
+    type: TypingEvents;
+  };
 }
 
 /**
--- a/src/core/room-options.ts
+++ b/src/core/room-options.ts
@@ -29,7 +29,7 @@ export const AllFeaturesEnabled = {
      *
      * Spec: CHA-T10.
      */
-    heartbeatThrottleMs: 15000,
+    heartbeatThrottleMs: 10000,
   } as TypingOptions,
 
   /**
--- a/src/core/room-options.ts
+++ b/src/core/room-options.ts
@@ -69,12 +69,13 @@ export interface PresenceOptions {
  */
 export interface TypingOptions {
   /**
-   * The time, in milliseconds, that a client will wait before emitting a `typing.started` event.
+   * A throttle, in milliseconds, that enforces the minimum time interval between consecutive `typing.started`
+   * events sent by the client to the server.
    * If typing.start() is called, the first call will emit an event immediately.
-   * Later calls will no-op until the interval has elapsed.
-   * Calling typing.stop() will immediately send a `typing.stopped` event and reset the interval,
+   * Later calls will no-op until the time has elapsed.
+   * Calling typing.stop() will immediately send a `typing.stopped` event to the server and reset the interval,
    * allowing the client to send another `typing.started` event immediately.
-   * @defaultValue 15000
+   * @defaultValue 10000
    */
   heartbeatThrottleMs: number;
 }
--- a/src/core/room-options.ts
+++ b/src/core/room-options.ts
@@ -25,29 +25,11 @@ export const AllFeaturesEnabled = {
    */
   typing: {
     /**
-     * The default time that a client will wait after calling typing.start() before emitting a `typing.stopped`
-     * event.
-     * Restarts the interval with repeated calls to typing.start(), resets the interval with typing.stop().
-     *
-     * Spec: CHA-T3
-     * @defaultValue
-     */
-    timeoutMs: 2000,
-
-    /**
      * The default time that a client will wait between sending one typing heartbeat and the next.
      *
      * Spec: CHA-T10.
      */
-    heartbeatIntervalMs: 15000,
-
-    /**
-     * The default timeout for typing inactivity in milliseconds.
-     *
-     * TODO: Rename this?
-     * Spec: CHA-T11
-     */
-    inactivityTimeoutMs: 500,
+    heartbeatThrottleMs: 15000,
   } as TypingOptions,
 
   /**
@@ -92,23 +74,9 @@ export interface TypingOptions {
    * Later calls will no-op until the interval has elapsed.
    * Calling typing.stop() will immediately send a `typing.stopped` event and reset the interval,
    * allowing the client to send another `typing.started` event immediately.
-   * @defaultValue 17000
+   * @defaultValue 15000
    */
-  heartbeatIntervalMs: number;
-
-  /**
-   * The optional timeout, in milliseconds, after which a client that pauses typing and does not resume, will emit a `typing.stopped` event.
-   * If not set, the client will not emit a `typing.stopped` event until they call `typing.stop()`.
-   * @defaultValue 2000
-   */
-  timeoutMs?: number | undefined;
-
-  /**
-   * The time, in milliseconds, a client waits after failing to receive a typing heartbeat from another client before assuming the other client has stopped typing.
-   * In practice, this means the client waits the length of the heartbeat interval plus this value before emitting a `typing.stopped` event.
-   * @defaultValue 500
-   */
-  inactivityTimeoutMs: number;
+  heartbeatThrottleMs: number;
 }
 
 /**
@@ -177,14 +145,8 @@ export const validateRoomOptions = (options: RoomOptions): void => {
 };
 
 const validateTypingOptions = (options: TypingOptions): void => {
-  if (options.timeoutMs !== undefined && options.timeoutMs <= 0) {
-    throw invalidRoomConfiguration('typing timeout must be greater than 0');
-  }
-  if (options.heartbeatIntervalMs <= 0) {
+  if (options.heartbeatThrottleMs <= 0) {
     throw invalidRoomConfiguration('typing heartbeat interval must be greater than 0');
-  }
-  if (options.inactivityTimeoutMs <= 0) {
-    throw invalidRoomConfiguration('typing inactivity timeout must be greater than 0');
   }
 };
 
--- a/src/core/room-options.ts
+++ b/src/core/room-options.ts
@@ -28,18 +28,26 @@ export const AllFeaturesEnabled = {
      * The default time that a client will wait after calling typing.start() before emitting a `typing.stopped`
      * event.
      * Restarts the interval with repeated calls to typing.start(), resets the interval with typing.stop().
+     *
+     * Spec: CHA-T3
+     * @defaultValue
      */
     timeoutMs: 2000,
 
     /**
      * The default time that a client will wait between sending one typing heartbeat and the next.
+     *
+     * Spec: CHA-T10.
      */
-    heartbeatIntervalMs: 17000,
+    heartbeatIntervalMs: 15000,
 
     /**
      * The default timeout for typing inactivity in milliseconds.
+     *
+     * TODO: Rename this?
+     * Spec: CHA-T11
      */
-    inactivityTimeoutMs: 2000,
+    inactivityTimeoutMs: 500,
   } as TypingOptions,
 
   /**
@@ -93,12 +101,12 @@ export interface TypingOptions {
    * If not set, the client will not emit a `typing.stopped` event until they call `typing.stop()`.
    * @defaultValue 2000
    */
-  timeoutMs?: number;
+  timeoutMs?: number | undefined;
 
   /**
    * The time, in milliseconds, a client waits after failing to receive a typing heartbeat from another client before assuming the other client has stopped typing.
    * In practice, this means the client waits the length of the heartbeat interval plus this value before emitting a `typing.stopped` event.
-   * @defaultValue 2000
+   * @defaultValue 500
    */
   inactivityTimeoutMs: number;
 }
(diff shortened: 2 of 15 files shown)


src/core/typing.ts
```diff
@@ -46,21 +43,24 @@ export interface Typing extends EmitsDiscontinuities {
...
-  get(): Promise<Set<string>>;
+  get(): Set<string>;
 
   /**
-   * Start indicates that the current user is typing. This will emit a typingStarted event to inform listening clients and begin a timer,
-   * once the timer expires, a typingStopped event will be emitted. The timeout is configurable through the typingTimeoutMs parameter.
-   * If the current user is already typing, it will reset the timer and being counting down again without emitting a new event.
+   * Start indicates that the current user is typing.
+   * This will emit a `typing.start` event to inform listening clients and begin a heartbeat timer,
+   * which can be configured through the `heartbeatIntervalMs` parameter.
```
AndyTWF, line 51: We probably don't need to describe the timer/internal workings in the public API, perhaps something as simple as:

"This will send a `typing.started` event to the server. Events are throttled according to the `heartbeatIntervalMs` room option. If an event has been sent within the interval, this operation is no-op."

src/core/typing.ts
```diff
@@ -46,21 +43,24 @@ export interface Typing extends EmitsDiscontinuities {
...
    */
 
   start(): Promise<void>;
 
   /**
-   * Stop indicates that the current user has stopped typing. This will emit a typingStopped event to inform listening clients,
-   * and immediately clear the typing timeout timer.
+   * Stop indicates that the current user has stopped typing.
+   * This will emit a `typing.stop` event to inform listening clients,
+   * and immediately clear any active timers.
```
AndyTWF, line 63:
```suggestion
   * Stop indicates that the current user has stopped typing.
   * This will emit a `typing.stopped` event to inform listening clients.
```

src/core/typing.ts
```diff
@@ -171,47 +167,77 @@ export class DefaultTyping
...
+
+    // CHA-T4d
+    if (this.channel.state !== 'attached' && this.channel.state !== 'attaching') {
+      this._logger.error(`DefaultTyping.start(); channel is not attached`, { state: this.channel.state });
+      throw new Ably.ErrorInfo('cannot start typing, channel is not attached', ErrorCodes.BadRequest, 400);
+    }
+
+    // If the user is already typing, and the timer has not expired, do not send another heartbeat
+    // CHA-T4c1, CHA-T4c2
+    if (this._heartbeatTimerId) {
```
vladvelici, line 199: Since this is just a throttle, wouldn't it be better to just save the last time this function was called and compare that with current timestamp, instead of actually using a timer?

---

When we set the throttle is also important. Right now this is an issue:
```
p1 = start() // fires publish
p2 = start() // fires publish, should've been throttled (in real use case: if user types faster than ack comes back; or react bad implementation)
// p1 resolves
// p2 resolves
p3 = start() // throttled
```

So the timer/throttle should be set when start() is called not when the publish finishes, otherwise we get double heartbeats as above.

We could choose to unset throttle if the publish call fails so we don't wait the full throttle period (although this should only happen if the call that fails is the call that set the throttle in the first place)?

Might have to do some thinking on how failing to publish should be handled as there can be strange edge cases, ie
```
p1 = start()
p2 = start()
// p2 throttled

// 100 ms later
p3 = start()
// p3 throttled

// another 50ms later
// p1 finishes but publish failed
```

src/core/typing.ts
```diff
@@ -171,47 +167,77 @@ export class DefaultTyping
...
+          this._heartbeatTimerId = undefined;
+        }
+      }, this._heartbeatIntervalMs));
+    }
   }
 
   /**
    * @inheritDoc
    */
   async start(): Promise<void> {
```
vladvelici, line 188: Should the return type tell the user if they've been throttled? Otherwise it'll be very strange experience that sometimes a request gets sent and sometimes it doesn't without any indication of whether you've actually sent a heartbeat or not.

src/core/events.ts
```diff
@@ -72,9 +72,44 @@ export enum PresenceEvents {
...
+}
+
+/**
+ * Represents a typing event payload.
+ */
+export interface TypingEvent {
+  /**
+   * Get a set of clientIds that are currently typing.
+   */
+  get currentlyTyping(): Set<string>;
```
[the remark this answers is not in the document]
splindsay-92: Yeh I think it can just be a set, I'll make the change.

src/core/room-options.ts
```diff
@@ -67,11 +69,14 @@ export interface PresenceOptions {
  */
 export interface TypingOptions {
   /**
-   * The timeout for typing events in milliseconds. If typing.start() is not called for this amount of time, a stop
-   * typing event will be fired, resulting in the user being removed from the currently typing set.
-   * @defaultValue 5000
+   * The time, in milliseconds, that a client will wait before emitting a `typing.started` event.
```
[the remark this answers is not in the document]
splindsay-92: I'll reword this, it's now a throttle too, so hopefully that makes it clearer.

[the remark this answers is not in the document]
splindsay-92: How about this
```ts
/**
   * A throttle, in milliseconds, that enforces the minimum time interval between consecutive `typing.started`
   * events sent by the client to the server.
   * If typing.start() is called, the first call will emit an event immediately.
   * Later calls will no-op until the time has elapsed.
   * Calling typing.stop() will immediately send a `typing.stopped` event to the server and reset the interval,
   * allowing the client to send another `typing.started` event immediately.
   * @defaultValue 10000
   */
   ```

File: src/core/typing.ts (hunk 4 above)
Comment (splindsay-92) [reply src/core/typing.ts:188]: I don't think so, the user doesn't have to know the internal logic here (though it should be pretty clear at a glance). Also, should a user (likely a dev) want to know why this didn't send a heartbeat, the log emitted should be a clear indication too.

Comment (vladvelici) [reply src/core/typing.ts:188]: I'd argue that not giving this indication is adding implicit behaviour that we wanted to stay away from.

In the case they want to make sure the indicator is on and stays on, they likely want to know the call succeeded and really sent a heartbeat. If they don't want it, they can ignore the value.

You can't react to logs, but you can react to a bool:
`ensureHeartbeat() { typing.start().then(sent => if (!sent) ensureHeartbeat()) }`

File: src/core/typing.ts (hunk 3 above)
Comment (splindsay-92) [reply src/core/typing.ts:199]: For me, I think a timer is still fine to use and a bit cleaner; we can clear it easily if there is an error in publish. Good spot on the missing catch though, I'll move things, so we start the time before we call publish, that should mean later calls no-op, and clear it on failure.

As for the last use-case, you make a good point. I think a mutex there that controls the flow of start/stops could work well and ensure consistent order for each function resolving.

File: src/core/typing.ts (hunk 4 above)
Comment (splindsay-92) [reply src/core/typing.ts:196]: I would say the user controls the implicit behavior through the throttle, reduce this to zero and you have full control over when you send a heartbeat.
Secondly, I think it could lead to confusion - you call `start()` but get a `false` may lead you to think the call has not succeeded. 

Also, the code above would call recursively until the heartbeat throttle expires, I'm don't think that's a good pattern to use. You could add a timer, but how will the user know what to set this too, without looking through the SDK and fully understanding how our implementation of heartbeats works.

Comment (vladvelici) [reply src/core/typing.ts:195]: > I would say the user controls the implicit behavior through the throttle, reduce this to zero and you have full control over when you send a heartbeat.

They can't control it with the throttle because a large chunk of how often heartbeats are expected from _others_ is controlled by this value too, so reducing it means you require more frequent heartbeats.

But we can just leave it and re-consider if we get feedback that someone actually needs to do something useful with that info.

Comment (splindsay-92) [reply src/core/typing.ts:195]: Happy to park for now then, but I'll add a note so we revisit this later and see if anything should change.

**Review by AndyTWF — APPROVED** [review]

File: src/core/typing.ts [hunk 7]
```diff
@@ -109,15 +104,18 @@ export class DefaultTyping
...
-  private _retryTimeout: ReturnType<typeof setTimeout> | undefined;
-  private _numRetries = 0;
+  // Grace period for inactivity before another user is considered to have stopped typing
+  // CHA-T10a
+  private readonly _timeoutMs = 2000;
+  private _heartbeatTimerId: ReturnType<typeof setTimeout> | undefined;
+  private readonly _currentlyTyping: Map<string, ReturnType<typeof setTimeout> | undefined>;
+
+  // Mutex for controlling `start` and `stop` operations
+  private readonly _mutex = new Mutex();
```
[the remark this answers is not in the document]
splindsay-92: Yeh we are trying to ensure order of execution is serial and predictable for users here :)

src/core/typing.ts
```diff
@@ -171,47 +174,94 @@ export class DefaultTyping
...
-    if (this._timerId) {
-      this._logger.debug(`DefaultTyping.start(); already typing, resetting timer`);
-      clearTimeout(this._timerId);
-      this._startTypingTimer();
+    // Acquire a mutex
+    await this._mutex.acquire();
+    // CHA-T4d
+    if (this.channel.state !== 'attached' && this.channel.state !== 'attaching') {
+      this._logger.error(`DefaultTyping.start(); channel is not attached`, { state: this.channel.state });
+      throw new Ably.ErrorInfo('cannot start typing, channel is not attached', ErrorCodes.BadRequest, 400);
```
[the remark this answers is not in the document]
splindsay-92: I've now re-written the function, so we correctly handle this too and use the try/catch/finally pattern throughout

src/core/typing.ts
```diff
@@ -238,82 +288,139 @@ export class DefaultTyping
...
-        if (!willReattempt) {
-          return;
-        }
+    // Stop typing: clear their timeout and remove from the currently typing set
+    this._logger.debug(`DefaultTyping._handleTypingStop(); client stopped typing`, { clientId });
+    clearTimeout(existingTimeout);
+    this._currentlyTyping.delete(clientId);
+    // Emit stop event only when the client is removed
+    this.emit(TypingEvents.Stop, {
+      currentlyTyping: new Set<string>(this._currentlyTyping.keys()),
```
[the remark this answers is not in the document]
splindsay-92: Currently, it's in a map as we need to easily access and update timers for each userId. I think just pulling the keys is a pretty quick operation, but happy to change this if there is a better way?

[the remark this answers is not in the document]
JoaoDiasAbly: Forgot `Map.keys()` is of type `IterableIterator`, so we do need to collect the keys in some collection 😅 disregard this comment

JoaoDiasAbly (approved):

src/core/typing.ts
```diff
@@ -348,4 +454,13 @@ export class DefaultTyping
   get detachmentErrorCode(): ErrorCodes {
     return ErrorCodes.TypingDetachmentFailed;
   }
+
+  // Convenience getters for testing
+  get heartbeatTimerId(): TypingTimerHandle {
```
vladvelici, line 459: These are visible in JS, maybe add a `_` as prefix if they have to be here?
Or better just cast it to another type in the test and directly use the _heartbeatTimerId (which already has a `_`)? Same for `currentlyTyping` if that's not intended as public API (doesn't look like since it returns timer IDs).

Not for this PR: I think we should look at how our SDK looks like in JavaScript (ie autocomplete suggestions in browser console) and try to make it look like the public API as much as possible. We have rather busy public classes. The interface are nice and simple, though. The same experience should be when working in JS directly (playing REPL in node or browser consoles).

src/core/typing.ts
```diff
@@ -238,82 +291,139 @@ export class DefaultTyping
...
-          clearTimeout(this._retryTimeout);
-          this._retryTimeout = undefined;
-          this._numRetries = 0;
-        }
+  /**
+   * Handles logic for TypingEvents.Start, including starting a new timeout or resetting an existing one.
+   * @param clientId
+   * @param existingTimeout
+   */
+  private _handleTypingStart(clientId: string, existingTimeout: NodeJS.Timeout | undefined): void {
```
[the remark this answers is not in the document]
splindsay-92: Sounds good to me, I've made the change :)

src/core/typing.ts
splindsay-92: For now, I've just created an interface in the test which exposes it that way, so that should help declutter this class a bit too.

src/core/events.ts
```diff
@@ -69,9 +69,44 @@ export enum PresenceEvents {
...
+  change: {
+    /**
+     * The client ID of the user who stopped/started typing.
+     */
+    clientId: string;
+
+    /**
+     * Type of the event.
+     */
+    type: TypingEvents;
```
ttypic, line 108: Sorry for the late comment, but is it a bit weird, that type of the event is not on the event object? Instead it's nested inside change? Wouldn't it be more consistent to have: 

```ts
export interface TypingEvent {
  clientId: string;
  type: TypingEvents;
  currentlyTyping: Set<string>;
}
```

src/core/typing.ts
```diff
@@ -171,47 +190,106 @@ export class DefaultTyping
...
+        return;
+      }
+      throw new Ably.ErrorInfo('mutex acquisition failed', 50000, 500);
+    });
+    try {
+      // CHA-T4d
+      // Ensure channel is attached
+      if (this.channel.state !== 'attached' && this.channel.state !== 'attaching') {
+        this._logger.error(`DefaultTyping.keystroke(); channel is not attached`, { state: this.channel.state });
+        throw new Ably.ErrorInfo('cannot type, channel is not attached', 50000, 500);
```
AndyTWF, line 228: This error message is a bit misleading - as it can be `attaching` too

src/core/typing.ts
```diff
@@ -44,23 +42,40 @@ export interface Typing extends EmitsDiscontinuities {
...
+   * If an event has been sent within the interval, this operation is no-op.
+   *
+   *
+   * Calls to `keystroke()` and `stop()` are serialized and will always resolve in the correct order.
+   * - For example, if multiple `keystroke()` calls are made in quick succession before the first `keystroke()` call has
+   *   sent a `typing.started` event to the server, followed by one `stop()` call, the `stop()` call will execute
+   *   as soon as the first `keystroke()` call completes.
+   *   All intermediate `keystroke()` calls will be treated as no-ops.
+   * - The most recent operation (`keystroke()` or `stop()`) will always determine the final state, ensuring operations
+   *   resolve to a consistent and correct state.
```
AndyTWF, line 61: This section should include multiple`@throws` tags to describe the error conditions

src/core/typing.ts
```diff
@@ -44,23 +42,40 @@ export interface Typing extends EmitsDiscontinuities {
...
+   * This will send a `typing.stopped` event to the server.
+   * If the user was not currently typing, this operation is no-op.
+   *
+   * Calls to `keystroke()` and `stop()` are serialized and will always resolve in the correct order.
+   * - For example, if multiple `keystroke()` calls are made in quick succession before the first `keystroke()` call has
+   *   sent a `typing.started` event to the server, followed by one `stop()` call, the `stop()` call will execute
+   *   as soon as the first `keystroke()` call completes.
+   *   All intermediate `keystroke()` calls will be treated as no-ops.
+   * - The most recent operation (`keystroke()` or `stop()`) will always determine the final state, ensuring operations
+   *   resolve to a consistent and correct state.
```
AndyTWF, line 78: This section should include multiple`@throws` tags to describe the error conditions

src/core/events.ts
splindsay-92: Hmm, it may just be a naming issue. Would it be better if we copied the `TypingEvents` enum and named it `TypingStatus`? Or perhaps renaming TypingEvent to `TypingSetChangeEvent`?

splindsay-92: We opted for the `change` field since we are including the whole set of typers. If anything, the `type` for this event should be `TypingSetChange`. The TypingEvent didn't build the whole set, it simply moved it into its new state.

ttypic: Got it! I was thinking about this object a bit differently. I saw it as representing a typing event (as stated in the jsdoc), with `currentlyTyping` as an additional context of event. The most important payload for me is `type` and `clientId`. 

I think we should at least update the documentation a little. Instead of `Represents a typing event payload`, maybe something like `Represents a typing state change event payload`.

splindsay-92: > The most important payload for me is type and clientId.
Perhaps you could explain this in more detail? For ably-js, I think idea is that most devs would likely just use the set and directly update the UI state with it. 

> I think we should at least update the documentation a little. Instead of Represents a typing event payload, maybe something like Represents a typing state change event payload.
How about.. Represents a change in the state of current typers?

ttypic: > How about.. Represents a change in the state of current typers?

Like it!

> Perhaps you could explain this in more detail? For ably-js, I think idea is that most devs would likely just use the set and directly update the UI state with it.

It was just my initial understanding of typing event, after you explained it, I got it, it makes sense, just think jsdoc is misleading a little bit.

test/core/typing.test.ts
```diff
@@ -15,389 +18,661 @@ interface TestContext {
...
-    // Ensure that the listener received the event
-    expect(receivedEvents).toHaveLength(1);
-    expect(receivedEvents[0]).toEqual({
-      currentlyTyping: new Set(['otherClient']),
-    });
+        expect(realtimeChannel.publish).toHaveBeenCalledTimes(2);
+        // Ensure that publish was called with typing.started only once
+        expect(realtimeChannel.publish).toHaveBeenCalledWith(startMessage);
+        // Ensure that publish was called with typing.stopped only once
+        expect(realtimeChannel.publish).toHaveBeenCalledWith(stopMessage);
```
vladvelici, line 258: Should these be be `toHaveBeenNthCalledWith()` and `toHaveBeenLastCalledWith()` to also ensure start was published before stop?

vladvelici (approved):

src/core/typing.ts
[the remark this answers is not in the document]
sacOO7: I agree mutex surely feels like an anti-pattern in js or single threaded environment.
In case of Kotlin, we adapted to single-threaded approach and avoided the use of mutex, and based on spec, it should cover all concurrent cases. 
Link to the file -> 
https://github.com/ably/ably-chat-kotlin/blob/0886eec9584e082a8bc5f221f344557ecc6ec078/chat-android/src/main/java/com/ably/chat/Typing.kt

test/core/typing.test.ts
```diff
@@ -15,389 +18,661 @@ interface TestContext {
...
-    });
+        // If stop is called, it should publish a leave message
+        vi.spyOn(realtimeChannel, 'publish').mockImplementation(async (): Promise<void> => {});
+        vi.spyOn(room.typing.channel, 'state', 'get').mockReturnValue('attached');
 
-    await waitForArrayLength(receivedEvents, 1);
-    expect(channel.presence.get).toBeCalledTimes(1);
+        // Start typing and then immediately stop typing
+        await room.typing.keystroke();
+        await room.typing.stop();
```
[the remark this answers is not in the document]
splindsay-92: There is a test a little further up which tests the order resolution guarantees separately, but I'll make the changes here anyway :)

src/core/typing.ts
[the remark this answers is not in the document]
AndyTWF: In JS at least, we have to account for the fact that promises may execute and resolve out-of-order, especially when we're talking about WS/Network calls

src/core/typing.ts
```diff
@@ -109,15 +125,18 @@ export class DefaultTyping
...
-  private _currentlyTyping: Set<string> = new Set<string>();
-  private _retryTimeout: ReturnType<typeof setTimeout> | undefined;
-  private _numRetries = 0;
+  // Grace period for inactivity before another user is considered to have stopped typing
+  // CHA-T10a
+  private readonly _timeoutMs = 2000;
+  private _heartbeatTimerId: TypingTimerHandle;
+  private readonly _currentlyTyping: Map<string, TypingTimerHandle>;
+
+  // Mutex for controlling `start` and `keystroke` operations
```
AndyTWF, line 138:
```suggestion
  // Mutex for controlling `keystroke` and `stop` operations
```

AndyTWF (approved):

PR #489, opened 2025-03-05
npm scripts update

vladvelici:
- **add npm run test:no-integration command - useful for quickly re-running unit tests**
- **Make formatting silent when running lint:fix so lint errors are highly visible**


package.json
```diff
@@ -34,7 +34,8 @@
     "demo:reload": "npm run build && cd demo && npm i file:../",
     "docs": "typedoc",
     "modulereport": "tsc --noEmit --esModuleInterop scripts/moduleReport.ts && esr scripts/moduleReport.ts",
-    "cspell": "cspell '{src,test}/**' './*.md'"
+    "cspell": "cspell '{src,test}/**' './*.md'",
+    "test:no-integration": "vitest run --exclude \"test/**/*.integration.test.ts?(x)\""
```
AndyTWF, line 38: Perhaps flip this on its head to be `test:unit`?

AndyTWF: One thing we could actually do here, we could define the different suites in `vitest.workspace.ts` and then reference them here

vladvelici: Renamed to test:unit and added `test:unit-chat` and `test:unit-react` to be able to run the different projects separately.

AndyTWF (approved):

PR #505, opened 2025-03-17
feature: chat on a single channel

AndyTWF:
### Context

[CHA-867]

### Description

This PR contains the reference implementation for the single-channel migration, in a nutshell.

- Moved all features to use a single channel, `<roomID>::$chat`.
- All API calls now use "v3" of the HTTP API.
- Room Reactions now use the `ephemeral` flag on messages.
- Refactor of `RoomOptions`
  - These are now entirely optional, with defaults provided.
  - Presence of an object in the options no longer controls the "on-ness" of features. All features are on by default. Features that incur an additional cost for the users have separate flags that disable that part of the feature, i.e. occupancy and presence.
- Discontinuity handling now moved up to the Room level.
- Removed feature-level channel access, now at Room level.
- Re-implementation of the Room Lifecycle Manager to greatly simplify how it runs to use just one channel.
- Removed "transient detachment" handling. 

### Checklist

* [ ] QA'd by the author.

Please test the entire SDK thoroughly.

[CHA-867]: https://ably.atlassian.net/browse/CHA-867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/channel-manager.ts
+++ b/src/core/channel-manager.ts
@@ -24,7 +24,7 @@ export class ChannelManager {
   }
 
   mergeOptions(merger: ChannelOptionsMerger): void {
-    this._logger.trace('ChannelManager.registerOptions();');
+    this._logger.trace('ChannelManager.mergeOptions();');
     if (this._resolvedChannel) {
       this._logger.error('channel options cannot be modified after the channel has been requested');
       throw new Ably.ErrorInfo('channel options cannot be modified after the channel has been requested', 40000, 400);
--- a/src/core/channel-manager.ts
+++ b/src/core/channel-manager.ts
@@ -1,5 +1,6 @@
 import * as Ably from 'ably';
 
+import { roomChannelName } from './channel.js';
 import { Logger } from './logger.js';
 import { DEFAULT_CHANNEL_OPTIONS, DEFAULT_CHANNEL_OPTIONS_REACT } from './version.js';
 
@@ -8,42 +9,46 @@ export type ChannelOptionsMerger = (options: Ably.ChannelOptions) => Ably.Channe
 export class ChannelManager {
   private readonly _realtime: Ably.Realtime;
   private readonly _logger: Logger;
-  private readonly _registeredOptions = new Map<string, Ably.ChannelOptions>();
-  private readonly _requestedChannels = new Set<string>();
-  private readonly _isReact;
+  private _registeredOptions: Ably.ChannelOptions;
+  private readonly _isReact: boolean;
+  private _resolvedChannel?: Ably.RealtimeChannel;
+  private readonly _channelId: string;
 
-  constructor(realtime: Ably.Realtime, logger: Logger, isReact: boolean) {
+  constructor(roomId: string, realtime: Ably.Realtime, logger: Logger, isReact: boolean) {
     logger.trace('ChannelManager();', { isReact });
     this._realtime = realtime;
     this._logger = logger;
     this._isReact = isReact;
+    this._registeredOptions = this._defaultChannelOptions();
+    this._channelId = roomChannelName(roomId);
   }
 
-  mergeOptions(channelName: string, merger: ChannelOptionsMerger): void {
-    this._logger.trace('ChannelManager.registerOptions();', { channelName });
-    if (this._requestedChannels.has(channelName)) {
-      this._logger.error('channel options cannot be modified after the channel has been requested', { channelName });
+  mergeOptions(merger: ChannelOptionsMerger): void {
+    this._logger.trace('ChannelManager.registerOptions();');
+    if (this._resolvedChannel) {
+      this._logger.error('channel options cannot be modified after the channel has been requested');
       throw new Ably.ErrorInfo('channel options cannot be modified after the channel has been requested', 40000, 400);
     }
 
-    const currentOpts = this._registeredOptions.get(channelName) ?? this._defaultChannelOptions();
-    this._registeredOptions.set(channelName, merger(currentOpts));
+    this._registeredOptions = merger(this._registeredOptions);
   }
 
-  get(channelName: string): Ably.RealtimeChannel {
-    this._logger.trace('ChannelManager.get();', { channelName });
-    this._requestedChannels.add(channelName);
-    return this._realtime.channels.get(
-      channelName,
-      this._registeredOptions.get(channelName) ?? this._defaultChannelOptions(),
-    );
+  get(): Ably.RealtimeChannel {
+    this._logger.trace('ChannelManager.get();');
+    if (!this._resolvedChannel) {
+      this._resolvedChannel = this._realtime.channels.get(this._channelId, this._registeredOptions);
+    }
+
+    return this._resolvedChannel;
   }
 
-  release(channelName: string): void {
-    this._logger.trace('ChannelManager.release();', { channelName });
-    this._requestedChannels.delete(channelName);
-    this._registeredOptions.delete(channelName);
-    this._realtime.channels.release(channelName);
+  release(): void {
+    this._logger.trace('ChannelManager.release();', { channelId: this._channelId });
+    if (!this._resolvedChannel) {
+      return;
+    }
+
+    this._realtime.channels.release(this._channelId);
   }
 
   private _defaultChannelOptions(): Ably.ChannelOptions {
--- a/src/core/errors.ts
+++ b/src/core/errors.ts
@@ -20,58 +20,6 @@ export enum ErrorCodes {
   MessageRejectedByModeration = 42213,
 
   /**
-   * The messages feature failed to attach.
-   */
-  MessagesAttachmentFailed = 102001,
-
-  /**
-   * The presence feature failed to attach.
-   */
-  PresenceAttachmentFailed = 102002,
-
-  /**
-   * The reactions feature failed to attach.
-   */
-  ReactionsAttachmentFailed = 102003,
-
-  /**
-   * The occupancy feature failed to attach.
-   */
-  OccupancyAttachmentFailed = 102004,
-
-  /**
-   * The typing feature failed to attach.
-   */
-  TypingAttachmentFailed = 102005,
-  // 102006 - 102049 reserved for future use for attachment errors
-
-  /**
-   * The messages feature failed to detach.
-   */
-  MessagesDetachmentFailed = 102050,
-
-  /**
-   * The presence feature failed to detach.
-   */
-  PresenceDetachmentFailed = 102051,
-
-  /**
-   * The reactions feature failed to detach.
-   */
-  ReactionsDetachmentFailed = 102052,
-
-  /**
-   * The occupancy feature failed to detach.
-   */
-  OccupancyDetachmentFailed = 102053,
-
-  /**
-   * The typing feature failed to detach.
-   */
-  TypingDetachmentFailed = 102054,
-  // 102055 - 102099 reserved for future use for detachment errors
-
-  /**
    * The room has experienced a discontinuity.
    */
   RoomDiscontinuity = 102100,
--- a/src/core/errors.ts
+++ b/src/core/errors.ts
@@ -42,16 +42,6 @@ export enum ErrorCodes {
   RoomIsReleased = 102103,
 
   /**
-   * Cannot perform operation because the previous operation failed.
-   */
-  PreviousOperationFailed = 102104,
-
-  /**
-   * An unknown error has happened in the room lifecycle.
-   */
-  RoomLifecycleError = 102105,
-
-  /**
    * Room was released before the operation could complete.
    */
   RoomReleasedBeforeOperationCompleted = 102106,
--- a/src/core/messages.ts
+++ b/src/core/messages.ts
@@ -1,22 +1,12 @@
 import * as Ably from 'ably';
 
-import { messagesChannelName } from './channel.js';
 import { ChannelManager } from './channel-manager.js';
 import { ChatApi } from './chat-api.js';
-import {
-  DiscontinuityListener,
-  EmitsDiscontinuities,
-  HandlesDiscontinuity,
-  newDiscontinuityEmitter,
-  OnDiscontinuitySubscriptionResponse,
-} from './discontinuity.js';
-import { ErrorCodes } from './errors.js';
 import { ChatMessageActions, MessageEvent, MessageEvents, RealtimeMessageNames } from './events.js';
 import { Logger } from './logger.js';
 import { DefaultMessage, Message, MessageHeaders, MessageMetadata, MessageOperationMetadata } from './message.js';
 import { parseMessage } from './message-parser.js';
 import { PaginatedResult } from './query.js';
-import { ContributesToRoomLifecycle } from './room-lifecycle-manager.js';
 import { Subscription } from './subscription.js';
 import EventEmitter, { wrap } from './utils/event-emitter.js';
 
@@ -166,6 +156,11 @@ export type MessageListener = (event: MessageEvent) => void;
 export interface MessageSubscriptionResponse extends Subscription {
   /**
    * Get the previous messages that were sent to the room before the listener was subscribed.
+   *
+   * If you are subscribed to and notified of a discontinuity event, this will reset the starting point of
+   * getPreviousMessages to the point at which a new series of continuity is started. When this happens, you should
+   * re-call getPreviousMessages to fill the gap.
+   *
    * @param params Options for the history query.
    * @returns A promise that resolves with the paginated result of messages, in newest-to-oldest order.
    */
@@ -178,7 +173,7 @@ export interface MessageSubscriptionResponse extends Subscription {
  *
  * Get an instance via {@link Room.messages}.
  */
-export interface Messages extends EmitsDiscontinuities {
+export interface Messages {
   /**
    * Subscribe to new messages in this chat room.
    * @param listener callback that will be called
@@ -250,19 +245,12 @@ export interface Messages extends EmitsDiscontinuities {
    * @returns A promise of the updated message.
    */
   update(message: Message, details?: OperationDetails): Promise<Message>;
-
-  /**
-   * Get the underlying Ably realtime channel used for the messages in this chat room.
-   *
-   * @returns The realtime channel.
-   */
-  get channel(): Ably.RealtimeChannel;
 }
 
 /**
  * @inheritDoc
  */
-export class DefaultMessages implements Messages, HandlesDiscontinuity, ContributesToRoomLifecycle {
+export class DefaultMessages implements Messages {
   private readonly _roomId: string;
   private readonly _channel: Ably.RealtimeChannel;
   private readonly _chatApi: ChatApi;
@@ -274,7 +262,6 @@ export class DefaultMessages implements Messages, HandlesDiscontinuity, Contribu
     }>
   >;
   private readonly _logger: Logger;
-  private readonly _discontinuityEmitter = newDiscontinuityEmitter();
   private readonly _emitter = new EventEmitter<MessageEventsMap>();
 
   /**
@@ -288,7 +275,7 @@ export class DefaultMessages implements Messages, HandlesDiscontinuity, Contribu
   constructor(roomId: string, channelManager: ChannelManager, chatApi: ChatApi, clientId: string, logger: Logger) {
     this._roomId = roomId;
 
-    this._channel = this._makeChannel(roomId, channelManager);
+    this._channel = this._makeChannel(channelManager);
 
     this._chatApi = chatApi;
     this._clientId = clientId;
@@ -299,8 +286,8 @@ export class DefaultMessages implements Messages, HandlesDiscontinuity, Contribu
   /**
    * Creates the realtime channel for messages.
    */
-  private _makeChannel(roomId: string, channelManager: ChannelManager): Ably.RealtimeChannel {
-    const channel = channelManager.get(messagesChannelName(roomId));
+  private _makeChannel(channelManager: ChannelManager): Ably.RealtimeChannel {
+    const channel = channelManager.get();
 
     // attachOnSubscribe is set to false in the default channel options, so this call cannot fail
     void channel.subscribe([RealtimeMessageNames.ChatMessage], this._processEvent.bind(this));
@@ -438,13 +425,6 @@ export class DefaultMessages implements Messages, HandlesDiscontinuity, Contribu
         }
       });
     });
-  }
-
-  /**
-   * @inheritdoc Messages
-   */
-  get channel(): Ably.RealtimeChannel {
-    return this._channel;
   }
 
   /**
@@ -608,42 +588,5 @@ export class DefaultMessages implements Messages, HandlesDiscontinuity, Contribu
     } catch (error: unknown) {
       this._logger.error(`failed to parse incoming message;`, { channelEventMessage, error: error as Ably.ErrorInfo });
     }
-  }
-
-  /**
-   * @inheritdoc HandlesDiscontinuity
-   */
-  discontinuityDetected(reason?: Ably.ErrorInfo): void {
-    this._logger.warn('Messages.discontinuityDetected();', { reason });
-    this._discontinuityEmitter.emit('discontinuity', reason);
-  }
-
-  /**
-   * @inheritdoc EmitsDiscontinuities
-   */
-  onDiscontinuity(listener: DiscontinuityListener): OnDiscontinuitySubscriptionResponse {
-    this._logger.trace('Messages.onDiscontinuity();');
-    const wrapped = wrap(listener);
-    this._discontinuityEmitter.on(wrapped);
-
-    return {
-      off: () => {
-        this._discontinuityEmitter.off(wrapped);
-      },
-    };
-  }
-
-  /**
-   * @inheritdoc ContributesToRoomLifecycle
-   */
-  get attachmentErrorCode(): ErrorCodes {
-    return ErrorCodes.MessagesAttachmentFailed;
-  }
-
-  /**
-   * @inheritdoc ContributesToRoomLifecycle
-   */
-  get detachmentErrorCode(): ErrorCodes {
-    return ErrorCodes.MessagesDetachmentFailed;
   }
 }
--- a/src/core/messages.ts
+++ b/src/core/messages.ts
@@ -1,6 +1,5 @@
 import * as Ably from 'ably';
 
-import { ChannelManager } from './channel-manager.js';
 import { ChatApi } from './chat-api.js';
 import { ChatMessageActions, MessageEvent, MessageEvents, RealtimeMessageNames } from './events.js';
 import { Logger } from './logger.js';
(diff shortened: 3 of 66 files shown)


src/core/channel-manager.ts
```diff
@@ -8,42 +9,46 @@ export type ChannelOptionsMerger = (options: Ably.ChannelOptions) => Ably.Channe
 export class ChannelManager {
   private readonly _realtime: Ably.Realtime;
   private readonly _logger: Logger;
-  private readonly _registeredOptions = new Map<string, Ably.ChannelOptions>();
-  private readonly _requestedChannels = new Set<string>();
-  private readonly _isReact;
+  private _registeredOptions: Ably.ChannelOptions;
+  private _isReact = false;
```
splindsay-92, line 13: Both this field and channelId can be readonly

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,389 @@
...
+    this._isFirstAttach = firstAttach;
   }
 
-  /**
-   * Ends the room lifecycle operation.
-   */
-  private _endLifecycleOperation(): void {
-    this._logger.debug('RoomLifecycleManager._endLifecycleOperation();');
-    this._operationInProgress = false;
+  testForceExplicitDetach(explicitDetach: boolean) {
```
splindsay-92, line 386: It doesn't look like this function is used?

src/core/room-options.ts
```diff
@@ -77,57 +47,85 @@ export interface TypingOptions {
...
-   * {@link AllFeaturesEnabled.occupancy} to enable occupancy with default options.
+   * The presence options for the room.
    */
-  occupancy?: OccupancyOptions;
+  presence?: PresenceOptions;
 }
 
+/**
+ * Represents the normalized typing options for a chat room, which makes every property required
+ * except for `timeoutMs`, which is optional.
```
splindsay-92, line 103: `timeoutMs` is no longer a typing config option

src/core/room-options.ts
```diff
@@ -77,57 +47,85 @@ export interface TypingOptions {
...
  * Represents the normalized options for a chat room.
  */
-export interface NormalizedRoomOptions extends RoomOptions {
+export interface InternalRoomOptions {
   /**
    * Are we running the client in a React environment?
    */
   isReactClient: boolean;
+
+  typing: InternalTypingOptions;
```
splindsay-92, line 126: Type docs for the interface members?

src/core/room-reactions.ts
```diff
@@ -140,16 +118,16 @@ export class DefaultRoomReactions
   constructor(roomId: string, channelManager: ChannelManager, clientId: string, logger: Logger) {
```
splindsay-92, line 118: RoomId is no longer needed in the constructor

src/core/typing.ts
```diff
@@ -143,14 +121,14 @@ export class DefaultTyping
    */
   constructor(
     roomId: string,
```
splindsay-92, line 123: Same here, roomId is no longer needed

src/core/typing.ts
```diff
@@ -446,37 +424,16 @@ export class DefaultTyping
...
     return this._heartbeatThrottleMs;
   }
 
-  /**
-   * @inheritdoc ContributesToRoomLifecycle
-   */
-  get attachmentErrorCode(): ErrorCodes {
-    return ErrorCodes.TypingAttachmentFailed;
+  // Convenience getters for testing
+  get heartbeatTimerId(): ReturnType<typeof setTimeout> | undefined {
```
splindsay-92, line 432: These have been removed, perhaps they got missed in the rebase?

splindsay-92 (commented):
Some minor comments, but there was an issue I spotted with presence when testing things out in the demo, will try to reproduce and add to the review.

splindsay-92:
Maybe the hook isn't registering the correct channel state change?
https://github.com/user-attachments/assets/d6230652-47a6-427f-b0cd-67aae73f473e

src/core/room-options.ts
AndyTWF: Fixed

src/core/room-options.ts
AndyTWF: Done

src/core/room-reactions.ts
AndyTWF: I've added the roomId to log messages as that's useful to have

src/core/typing.ts
AndyTWF: Ditto

src/core/errors.ts
```diff
@@ -19,58 +19,6 @@ export enum ErrorCodes {
...
-   */
-  OccupancyDetachmentFailed = 102053,
-
-  /**
-   * The typing feature failed to detach.
-   */
-  TypingDetachmentFailed = 102054,
-  // 102055 - 102099 reserved for future use for detachment errors
-
   /**
```
splindsay-92, line 22: Are `PreviousOperationFailed` and `RoomLifecycleError` errors still needed?

src/core/channel-manager.ts
vladvelici, line 1: Do we still need the `ChannelManager` abstraction?

Might be simpler to just pass a channel instance to feature classes and implement the channel options settings in the Room object (can still use the `ChannelOptionsMerger` approach for that)?

src/core/typing.ts
AndyTWF: Done!

src/core/channel-manager.ts
AndyTWF: We could get rid of it - I just thought it might be a nice way to keep the management of channels away from the room itself. Plus it does give us a foot-gun prevention in that if you try to say "I want different options" after the channel is made, it tells you that they won't have any effect. I'm happy either way, WDYT?

vladvelici: I'm not very opinionated either way, but [I have fallen into the trap of potentially setting options too late](https://github.com/ably/ably-chat-js/pull/473#discussion_r2005521158). The _room_ way and passing the channel to features mean that we can't fall into that trap.

> Plus it does give us a foot-gun prevention in that if you try to say "I want different options" after the channel is made

If we make the channel once at room level and pass it down to features, there's no foot-gun, channel is only created once per room?

AndyTWF: True, ok I'll look at removing the ChannelManager this afternoon

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,394 @@
...
-      this._releaseInProgress = true;
-      this._lifecycle.setStatus({ status: RoomStatus.Releasing });
+  private _checkRoomNotReleasing(op: string) {
+    switch (this._roomLifecycle.status) {
+      case RoomStatus.Released: {
+        throw new Ably.ErrorInfo(`cannot ${op} room, room is released`, ErrorCodes.RoomIsReleased, 400);
+      }
+      case RoomStatus.Releasing: {
+        throw new Ably.ErrorInfo(`cannot ${op} room, room is currently releasing`, ErrorCodes.RoomIsReleasing, 400);
+      }
```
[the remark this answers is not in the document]
sacOO7: We should update both code and the related spec.
Also, all in progress checks `ATTACHING`, `DETACHING`, `RELEASING` etc should be removed, since we only return when the channel reaches some deterministic state.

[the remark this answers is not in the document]
sacOO7: Okay, need to check for other states `ATTACHING`, `DETACHING` but `RELEASING` will always return false.

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,394 @@
...
-  constructor(
-    lifecycle: DefaultRoomLifecycle,
-    contributors: ContributesToRoomLifecycle[],
-    logger: Logger,
-    transientDetachTimeout: number,
-  ) {
+  private readonly _roomId: string;
+  private readonly _eventEmitter: EventEmitter<RoomLifeCycleEvents>;
+  private _isFirstAttach: boolean; // CHA-RL13
+  private _isPostExplicitDetach: boolean; // CHA-RL14
```
sacOO7, line 39:
```suggestion
  private _isExplicitlyDetached: boolean; // CHA-RL14
```

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,394 @@
...
-   */
-  constructor(
-    lifecycle: DefaultRoomLifecycle,
-    contributors: ContributesToRoomLifecycle[],
-    logger: Logger,
-    transientDetachTimeout: number,
-  ) {
+  private readonly _roomId: string;
+  private readonly _eventEmitter: EventEmitter<RoomLifeCycleEvents>;
+  private _isFirstAttach: boolean; // CHA-RL13
```
sacOO7, line 38:
```suggestion
  private _isAttachedOnce: boolean; // CHA-RL13
```
Initial value will be `false`

sacOO7: Since, variable name is explicit, there's no guessing as to whether it's set before first ATTACH or after first ATTACH.

src/core/room-lifecycle-manager.ts
sacOO7: New name makes it clear that it's detached by user manually and flag is when the detach operation completes.

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,394 @@
...
+    this._roomLifecycle = roomLifecycle;
     this._logger = logger;
-    this._contributors = contributors;
-    this._transientDetachTimeouts = new Map();
-    this._lifecycle = lifecycle;
+    this._eventEmitter = new EventEmitter();
+    this._isFirstAttach = true; // CHA-RL13
+    this._isPostExplicitDetach = false; // CHA-RL14
+    this._mutex = new Mutex(); // CHA-RL7
+    this._operationInProgress = false;
```
sacOO7, line 52:
```suggestion
    this._mutex = new Mutex(); // CHA-RL7
    // this._operationInProgress not needed
```


`_operationInProgress` indicates whether `this._mutex` has an operation in progress.
I firmly believe check for the same can be replaced by `this._mutex.isLocked()`, [doc link](https://github.com/DirtyHairy/async-mutex/tree/master?tab=readme-ov-file#checking-whether-the-mutex-is-locked)
Since JS is single-threaded, it should result in exact same behaviour.

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,394 @@
...
-    contributors: ContributesToRoomLifecycle[],
-    logger: Logger,
-    transientDetachTimeout: number,
-  ) {
+  private readonly _roomId: string;
+  private readonly _eventEmitter: EventEmitter<RoomLifeCycleEvents>;
+  private _isFirstAttach: boolean; // CHA-RL13
+  private _isPostExplicitDetach: boolean; // CHA-RL14
+  private readonly _mutex: Mutex; // CHA-RL7
+  private _operationInProgress: boolean;
```
sacOO7, line 41:
```suggestion
  private readonly _mutex: Mutex; // CHA-RL7
  // private _operationInProgress: boolean; - Not needed
```

`_operationInProgress` indicates whether `_mutex` has an operation in progress.
I firmly believe check for the same can be replaced by `_mutex.isLocked()`, [doc link](https://github.com/DirtyHairy/async-mutex/tree/master?tab=readme-ov-file#checking-whether-the-mutex-is-locked)
Since JS is single-threaded, it should result in exact same behaviour.
This way, we don't need to worry about setting/resetting `_operationInProgress` at the end of each mtx operation.

AndyTWF: [d004110](https://github.com/ably/ably-chat-js/pull/505/commits/d0041104aa425104e931b758c8649ba11987bb27)

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,394 @@
...
-      .runExclusive(() => {
-        this._logger.error('RoomLifecycleManager._onChannelSuspension(); setting room status to contributor status', {
-          status: contributor.channel.state as RoomStatus,
-          error: detachError,
-        });
-        this._lifecycle.setStatus({
-          status: contributor.channel.state as RoomStatus,
-          error: detachError,
+      // CHA-RL11b
+      if (this._operationInProgress) {
```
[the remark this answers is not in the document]
AndyTWF: [d004110](https://github.com/ably/ably-chat-js/pull/505/commits/d0041104aa425104e931b758c8649ba11987bb27)

src/core/room-lifecycle-manager.ts
AndyTWF: [d004110](https://github.com/ably/ably-chat-js/pull/505/commits/d0041104aa425104e931b758c8649ba11987bb27)

src/core/room-lifecycle-manager.ts
AndyTWF: [d004110](https://github.com/ably/ably-chat-js/pull/505/commits/d0041104aa425104e931b758c8649ba11987bb27)

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,392 @@
...
+  private _roomStatusIs(status: RoomStatus) {
+    return this._roomLifecycle.status === status;
+  }
 
-      // We force the room status to be releasing
-      this._clearAllTransientDetachTimeouts();
-      this._startLifecycleOperation();
-      this._releaseInProgress = true;
-      this._lifecycle.setStatus({ status: RoomStatus.Releasing });
+  private async _channelDetachLoop(channel: Ably.RealtimeChannel) {
```
sacOO7, line 345:
```suggestion
  private async _retryUntilChannelDetached(channel: Ably.RealtimeChannel) {
```

src/core/room.ts
```diff
@@ -114,20 +115,35 @@ export interface Room {
    * @returns A copy of the options used to create the room.
    */
   options(): RoomOptions;
+
```
splindsay-92, line 118: The above accessors still state that we will throw an error if a feature is not enabled, but this is no longer the case.
e.g.
```
/**
   * Allows you to interact with room-level reactions.
   *
   * @throws {@link ErrorInfo} if reactions are not enabled for the room.
   * @returns The room reactions instance for the room.
   */
  get reactions(): RoomReactions;
```

src/core/channel-manager.ts
```diff
@@ -8,42 +9,46 @@ export type ChannelOptionsMerger = (options: Ably.ChannelOptions) => Ably.Channe
...
+    this._registeredOptions = this._defaultChannelOptions();
+    this._channelId = roomChannelName(roomId);
   }
 
-  mergeOptions(channelName: string, merger: ChannelOptionsMerger): void {
-    this._logger.trace('ChannelManager.registerOptions();', { channelName });
-    if (this._requestedChannels.has(channelName)) {
-      this._logger.error('channel options cannot be modified after the channel has been requested', { channelName });
+  mergeOptions(merger: ChannelOptionsMerger): void {
+    this._logger.trace('ChannelManager.registerOptions();');
```
splindsay-92, line 27: Should the log statement be `'ChannelManager.mergeOptions();'`?

src/core/errors.ts
splindsay-92: I think these can probably be removed now :)

src/core/messages.ts
```diff
@@ -167,6 +155,11 @@ export type MessageListener = (event: MessageEvent) => void;
 export interface MessageSubscriptionResponse extends Subscription {
   /**
    * Get the previous messages that were sent to the room before the listener was subscribed.
+   *
+   * If you are subscribed to and notified of a discontinuity event, this will reset the starting point of
```
splindsay-92, line 159: Might just be me reading this, but it sort of reads like you have to be subscribed and notified of a discontinuity for the resetting here to take place. This happens regardless of if you have a listener subscribed to discontinuity events though.

What about something like..
```suggestion
* If the client experiences a period of discontinuity, the `getPreviousMessages` starting point will be reset.
* Calls to `getPreviousMessages` during this period will wait for discontinuity to be restored before resolving.
* When continuity is restored, the starting point will be set to the beginning of this new period of continuity.
* As such, to ensure no messages are missing, you should re-call `getPreviousMessages` after any period of discontinuity.
```

src/core/room.ts
```diff
@@ -377,7 +347,32 @@ export class DefaultRoom implements Room {
     return this._lifecycle;
   }
 
+  /**
+   * @internal
+   */
   get channelManager(): ChannelManager {
```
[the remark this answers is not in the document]
AndyTWF: Nope, it can go!

src/core/errors.ts
AndyTWF: 5b22df6364ebf2886f7a56a4cc250a40f3f59922

src/core/room.ts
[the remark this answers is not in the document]
AndyTWF: 5b22df6364ebf2886f7a56a4cc250a40f3f59922

src/core/room.ts
AndyTWF: 5b22df6364ebf2886f7a56a4cc250a40f3f59922

src/core/messages.ts
AndyTWF: 5b22df6364ebf2886f7a56a4cc250a40f3f59922

src/core/occupancy.ts
```diff
@@ -199,49 +171,27 @@ export class DefaultOccupancy
       return;
     }
 
+    this._logger.debug('emitting occupancy update');
```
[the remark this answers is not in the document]
AndyTWF: adad12f7550bf0218713736963d7f6af0ec633b4

src/core/presence.ts
```diff
@@ -163,60 +153,40 @@ export interface Presence extends EmitsDiscontinuities {
...
    * Constructs a new `DefaultPresence` instance.
    * @param roomId The unique identifier of the room.
-   * @param channelManager The channel manager to use for creating the presence channel.
+   * @param channel The Realtime channel instance.
    * @param clientId The client ID, attached to presences messages as an identifier of the sender.
    * A channel can have multiple connections using the same clientId.
    * @param logger An instance of the Logger.
    */
-  constructor(roomId: string, channelManager: ChannelManager, clientId: string, logger: Logger) {
+  constructor(roomId: string, channel: Ably.RealtimeChannel, clientId: string, logger: Logger) {
```
splindsay-92, line 174: Doesn't look like roomId is used

AndyTWF: 6c56c17

src/core/presence.ts
```diff
@@ -361,34 +331,14 @@ export class DefaultPresence
...
   /**
-   * @inheritDoc
+   * Returns the channel name for the presence channel.
+   *
+   * @param roomId The unique identifier of the room.
+   * @returns The channel name for the presence channel.
    */
-  get detachmentErrorCode(): ErrorCodes {
-    return ErrorCodes.PresenceDetachmentFailed;
+  static channelName(roomId: string): string {
```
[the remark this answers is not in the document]
AndyTWF: 6c56c17

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,392 @@
...
-          channel: contributor.channel.name,
-          change,
-        });
-        contributor.discontinuityDetected(change.reason);
+  private _startMonitoringChannelState(): void {
+    const channel = this._channelManager.get();
+
+    // CHA-RL11a
+    channel.on((stateChange: Ably.ChannelStateChange) => {
+      this._logger.debug('channel state changed', {
```
splindsay-92, line 67:
```suggestion
      this._logger.debug('_startMonitoringChannelState();channel state changed', {
```

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,392 @@
...
-        this._logger.error('RoomLifecycleManager._onChannelSuspension(); setting room status to contributor status', {
-          status: contributor.channel.state as RoomStatus,
-          error: detachError,
-        });
-        this._lifecycle.setStatus({
-          status: contributor.channel.state as RoomStatus,
-          error: detachError,
+      // CHA-RL11b
+      if (this._operationInProgress()) {
+        this._logger.debug('ignoring channel state change - operation in progress', {
```
splindsay-92, line 77:
```suggestion
        this._logger.debug('_startMonitoringChannelState();ignoring channel state change - operation in progress', {
```

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,392 @@
...
-        if (this._lifecycle.status === RoomStatus.Failed) {
-          throw new Error('room is in a failed state');
-        }
-
-        // If not, we wait a short period and then try again
-        return new Promise<unknown>((resolve) => {
-          setTimeout(() => {
-            resolve(this._doChannelWindDown(contributor));
-          }, 250);
+        this._logger.warn('discontinuity detected', {
```
splindsay-92, line 109:
```suggestion
        this._logger.warn('_startMonitoringDiscontinuity(); discontinuity detected', {
```

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,392 @@
...
-          // As every other channel is wound down.
-          this._endLifecycleOperation();
-          throw change.reason ?? new Ably.ErrorInfo('unknown error in _doRetry', ErrorCodes.RoomLifecycleError, 500);
-        }
-      };
-      contributor.channel.on(listener);
-    }).then(() => {
-      this._logger.debug('RoomLifecycleManager._doRetry(); feature reattached via listener, retrying attach');
-      return doAttachWithRetry();
+        this._logger.warn('discontinuity detected', {
```
splindsay-92, line 133: Looks like a few logs in this file don't follow the standard format
```suggestion
        this._logger.warn('_startMonitoringDiscontinuity(); discontinuity detected', {
```

AndyTWF: [6b7eb10](https://github.com/ably/ably-chat-js/pull/505/commits/6b7eb10edf6786264adf942557fc7e8a17d44d36)

src/core/room-lifecycle-manager.ts
AndyTWF: [6b7eb10](https://github.com/ably/ably-chat-js/pull/505/commits/6b7eb10edf6786264adf942557fc7e8a17d44d36)

src/core/room-lifecycle-manager.ts
AndyTWF: [6b7eb10](https://github.com/ably/ably-chat-js/pull/505/commits/6b7eb10edf6786264adf942557fc7e8a17d44d36)

src/core/room-lifecycle-manager.ts
AndyTWF: [6b7eb10](https://github.com/ably/ably-chat-js/pull/505/commits/6b7eb10edf6786264adf942557fc7e8a17d44d36)

splindsay-92 (approved):
LGTM, thanks for making the changes!

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,410 @@
...
-          return;
-        }
-
-        // If we're not ignoring contributor detachments, we should process the event
-        this._logger.debug('RoomLifecycleManager(); update event received', {
-          channel: contributor.channel.name,
-          change,
-        });
-        contributor.discontinuityDetected(change.reason);
+  private _startMonitoringChannelState(): void {
```
sacOO7, line 62: Since, anyways we are listening to channel state changes, we can avoid duplicated code and listeners for `_startMonitoringDiscontinuity` as per 

```
  private _startMonitoringChannelState(): void {
    const channel = this._channelManager.get();

    // CHA-RL11a
    channel.on((stateChange: Ably.ChannelStateChange) => {
      this._logger.debug('RoomLifecycleManager.channel state changed', {
        roomId: this._roomId,
        oldState: stateChange.previous,
        newState: stateChange.current,
        reason: stateChange.reason,
        resumed: stateChange.resumed,
      });

      // CHA-RL11b, CHA-RL11c
      if (!this._operationInProgress()) {
        const newStatus = this._mapChannelStateToRoomStatus(stateChange.current);
        this._setStatus(newStatus, stateChange.reason);
      }

      if (channel.state == "attached") {
        if (!stateChange.resumed && this._hasAttachedOnce && !this._isExplicitlyDetached) {
          const error = new Ably.ErrorInfo(
            'discontinuity detected',
            ErrorCodes.RoomDiscontinuity,
            stateChange.reason?.statusCode ?? 0,
            stateChange.reason,
          );
  
          this._logger.warn('RoomLifecycleManager._startMonitoringChannelState(); discontinuity detected', {
            roomId: this._roomId,
            error,
          });
          this._eventEmitter.emit(RoomEvents.Discontinuity, error);
        }
      }
    });
  }
```

src/core/room-lifecycle-manager.ts
```diff
@@ -1,923 +1,410 @@
...
-            feature: failedFeature.channel.name,
-          });
-          return this._doRetry(failedFeature).catch();
-        }
-
-        // We attached, huzzah! It's the end of the loop
-        this._endLifecycleOperation();
-      });
-    };
+  private _startMonitoringDiscontinuity(): void {
```
sacOO7, line 99: We don't need this method as per https://github.com/ably/ably-chat-js/pull/505/files#r2034471154.
This method has duplicated code and two additional listeners.

src/core/room-lifecycle-manager.ts
sacOO7: This should result in exact same behaviour

sacOO7: wdyt

AndyTWF: I actually quite like the differentiation - it keeps the methods to handling one case only, rather than multiple things in play

src/core/channel-manager.ts
```diff
@@ -8,42 +9,46 @@ export type ChannelOptionsMerger = (options: Ably.ChannelOptions) => Ably.Channe
 export class ChannelManager {
```
sacOO7, line 9: Do we really need this class? Previously, we needed it to handle complex logic.
But things stand as of now, simple logic inside `src/core/channel-manager.ts` can be moved into `room.ts` class.
wdyt

AndyTWF: I thought about it for a while, in the end decided it was a nice encapsulation: https://github.com/ably/ably-chat-js/pull/505#discussion_r2013699894

sacOO7: This means we can directly inject a `room` instance into each feature. It also simplifies many things related to DI.

AndyTWF: It's something we can change in the future if it becomes a problem, I think - I don't see a strong argument for changing things at the moment

src/core/presence.ts
```diff
@@ -358,65 +330,24 @@ export class DefaultPresence implements Presence, HandlesDiscontinuity, Contribu
...
-      if (roomOptions.presence?.subscribe === undefined || roomOptions.presence.subscribe) {
-        channelModes.push('PRESENCE_SUBSCRIBE');
-      }
-
-      return { ...options, modes: channelModes };
+      const modes = options.modes ?? ['PUBLISH', 'SUBSCRIBE', 'PRESENCE'];
+      return {
+        ...options,
+        modes,
+      };
```
vladvelici, line 350: This part seems a bit weird to me.

1. Why does the presence options merger have default values that have nothing to do with presence?
2. Why does it mean that if we want the messages we do nothing (shouldn't then we set `PRESENCE` or `PRESENCE_SUBSCRIBE` as channel mode?
3. If the `PUBLISH, SUBSCRIBE, PRESENCE` should be there, shouldn't this merger actually check if the array has them and if not add them, as opposed to just checking if the array exists or not? It might as well exist but be empty?

src/react/hooks/use-typing.ts
```diff
@@ -92,14 +92,13 @@ export const useTyping = (params?: TypingParams): UseTypingResponse => {
...
       .then((room) => {
-        if (!mounted) return;
-
         // If we're not attached, we can't call typing.get() right now
         if (room.status === RoomStatus.Attached) {
           const typing = room.typing.get();
           logger.debug('useTyping(); room attached, getting initial typers', { typing });
           setCurrentlyTyping(typing);
         } else {
+          if (!mounted) return;
```
vladvelici, line 101: Was this move intentional? Do we want to call `setCurrentlyTyping` even if we're not still mounted?

AndyTWF: Nope, the code block is now redundant anyway, so removed in 5f8556bad0897b3fbf0561b094964188f92185dc

src/core/presence.ts
AndyTWF: > This part seems a bit weird to me.
> 
> 1. Why does the presence options merger have default values that have nothing to do with presence?
> 2. Why does it mean that if we want the messages we do nothing (shouldn't then we set `PRESENCE` or `PRESENCE_SUBSCRIBE` as channel mode?
> 3. If the `PUBLISH, SUBSCRIBE, PRESENCE` should be there, shouldn't this merger actually check if the array has them and if not add them, as opposed to just checking if the array exists or not? It might as well exist but be empty?

On 1. - no particular reason, just couldn't think of a better place to put it for now!

2. By default `ChannelOptions.modes` is `undefined` which then causes the channel to take whatever modes the server says it gets, which includes presence. So in this sense, if we _do_ want presence and presence_subscribe, we would just do nothing as they'll be given to us by default.

3. I tried to keep this method scoped to "just handle presence" - so left it doing nothing if nothing needed to be done.

I think this'll get a refactor in the future, but wanted to keep it simple for now

src/react/hooks/use-typing.ts
```diff
@@ -88,20 +88,13 @@ export const useTyping = (params?: TypingParams): UseTypingResponse => {
...
         if (room.status === RoomStatus.Attached) {
           const typing = room.typing.get();
           logger.debug('useTyping(); room attached, getting initial typers', { typing });
           setCurrentlyTyping(typing);
-        } else {
-          logger.debug('useTyping(); room not attached, setting currentlyTyping to empty');
-          setCurrentlyTyping(new Set());
         }
       })
       .catch();
```
vladvelici, line 100: I think these should go inside the wrapRoomPromise below instead of doing their own waiting? So the `unmount()` callback works correctly for those too?

It's likely this code isn't doing much though because at the time of mounting (at least on page load) the room very likely isn't yet attached. This block would make more sense in an effect that depends on room status.

It can be fixed in another PR, don't mean to drag this along. Maybe a ticket so we get back to it?

cc @splindsay-92 because it's for typing indicators.

vladvelici (approved):

src/react/hooks/use-typing.ts
AndyTWF: Yes, I agree - I've moved it into the same block.

PR #506, opened 2025-03-18
Fix: Encode `fromSerial`

splindsay-92:
### Context

* Ensure we are encoding the `fromSerial` parameter when it is provided in get requests.

### Checklist

* [x] QA'd by the author.

### DIFF
--- a/src/core/chat-api.ts
+++ b/src/core/chat-api.ts
@@ -103,6 +103,10 @@ export class ChatApi {
 
     // convert the params into internal format
     const apiParams: ApiGetMessagesQueryParams = { ...params };
+    // encode the serial if it is provided
+    if (params.fromSerial) {
+      apiParams.fromSerial = encodeURIComponent(params.fromSerial);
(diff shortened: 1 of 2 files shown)


src/core/chat-api.ts
```diff
@@ -103,6 +103,10 @@ export class ChatApi {
 
     // convert the params into internal format
     const apiParams: ApiGetMessagesQueryParams = { ...params };
+    // encode the serial if it is provided
+    if (params.fromSerial) {
+      apiParams.fromSerial = encodeURIComponent(params.fromSerial);
```
vladvelici, line 108: These are already encoded by `realtime.request`, no need to do it here.

I think doing it here will actually double encode and break decoding?

AndyTWF: Yeah I think what we have already is fine - we don't do anything special with it for `getPreviousMessages`, which uses the same mechanism and that works fine.

PR #507, opened 2025-03-20
dev: create cursor and copilot guidance

AndyTWF:
### Context

[CHA-879]

### Description

This PR adds cursor and copilot guidance files that provide project and Ably-specific information that helps the agents to better follow our development practices.

### Checklist

* [x] QA'd by the author.

N/A

[CHA-879]: https://ably.atlassian.net/browse/CHA-879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/copilot-instructions.md
```diff
@@ -0,0 +1,120 @@
...
+- Use relative imports
+- For Ably package: `import * as Ably from 'ably'`
+- Use named exports over default exports
+- Group related exports together
+- Export interfaces and types that are part of the public API in relevant index.ts files
+
+### Interface and Type Declarations
+- Use PascalCase for interface names
+- Use JSDoc style comments for interfaces
+- Document all properties using `@property`
```
vladvelici, line 25: Do we prefer properties to be defined at the interface-level or next to the property with `@property`? The style we use now seems to be this:

```
/**
 * Represents the detail of a message deletion or update.
 */
export interface Operation {
  /**
   * The optional clientId of the user who performed the update or deletion.
   */
  clientId?: string;
  /**
   * The optional description for the update or deletion.
   */
  description?: string;
  /**
   * The optional metadata associated with the update or deletion.
   */
  metadata?: MessageOperationMetadata;
}
```

When I ask copilot to generate docs for me it annoyingly puts them at the top of the interface/class (even without this instruction)

vladvelici (approved):
I'm approving this as-is.

We're better with than without. I think we should keep an eye on what AI does and see if we can tweak it as we go.

.github/copilot-instructions.md
vladvelici, line 1: Since this is much smaller than the cursor rules, would it be more useful to just have an automated script that merges all cursor rules into one file and let copilot read that?

Probably better results to have two but not sure if worth the cost of maintaining two sets of documentation for AI.

AndyTWF: So we'll definitely need two files because Copilot won't automatically read the cursor rules, but I'll see what I can do!

.github/copilot-instructions.md
AndyTWF: Yes I agree, fixed this.

.github/copilot-instructions.md
AndyTWF: I've added a script that combine the files from Cursor into one for GitHub automatically.

From some preliminary testing, it seems to be able to answer questions from it - so lets see how it goes!

PR #509, opened 2025-03-25
Refactor occupancy logging messages

splindsay-92:
### Description

* Standardised occupancy logs
* Corrected some incorrect docs

### Checklist

* [x] QA'd by the author.


src/core/occupancy.ts
```diff
@@ -165,37 +165,56 @@ export class DefaultOccupancy
    * occupancy events for the public API.
    */
   private _internalOccupancyListener(message: Ably.InboundMessage): void {
+    this._logger.trace('Occupancy._internalOccupancyListener();', message);
     if (typeof message.data !== 'object') {
-      this._logger.error('invalid occupancy event received; data is not an object', message);
+      this._logger.debug(
```
AndyTWF, line 170: Why would this not be an `error`? If we've setup the listeners correctly, are only expecting valid occupancy events, and we're receiving stuff we shouldn't be receiving, that seems like an error-level log to me?

splindsay-92: Ah, my bad, that shouldn't have gone in the commit!

AndyTWF (approved):

PR #518, opened 2025-04-03
Add unique listeners so unsubscribe() and stop extending EventEmitter

vladvelici:
Stop extending EventEmitter so we get a cleaner public API (in JS).

Wrap all listeners into a unique reference function to make sure that subscribing the same function twice and unsubscribing one subscription will only unsubscribes that one subscription and not both.

### Context

* [CHA-856]
* [CHA-506]

### Checklist

* [x] QA'd by the author.

[CHA-856]: https://ably.atlassian.net/browse/CHA-856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CHA-506]: https://ably.atlassian.net/browse/CHA-506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/room-status.ts
+++ b/src/core/room-status.ts
@@ -2,7 +2,7 @@ import * as Ably from 'ably';
 
 import { Logger } from './logger.js';
 import { StatusSubscription } from './subscription.js';
-import EventEmitter from './utils/event-emitter.js';
+import EventEmitter, { wrap } from './utils/event-emitter.js';
 
 /**
  * The different states that a room can be in throughout its lifecycle.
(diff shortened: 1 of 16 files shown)


splindsay-92 (commented):
Nice work! Do you think it's worth adding a test to cover this though? :)

src/core/utils/event-emitter.ts
```diff
@@ -60,3 +60,12 @@ const InternalEventEmitter: new <EventsMap>() => InterfaceEventEmitter<EventsMap
 class EventEmitter<EventsMap> extends InternalEventEmitter<EventsMap> {}
 
 export default EventEmitter;
+
+/**
+ * Creates a wrapper function that forwards all arguments to the provided function.
+ * @param fn The function to wrap
+ * @returns A new function with the same signature as the input function
+ */
+export function wrap<Args extends unknown[], Return>(fn: (...args: Args) => Return): (...args: Args) => Return {
```
AndyTWF, line 69: Nitpick: can we do this in the other format, e.g.

`export const wrap = (...) => ...`

To be consistent with elsewhere?

src/core/room-status.ts
```diff
@@ -194,11 +193,12 @@ export class DefaultRoomLifecycle extends EventEmitter<RoomStatusEventsMap> impl
    * @inheritdoc
    */
   onChange(listener: RoomStatusListener): StatusSubscription {
-    this.on(listener);
+    listener = wrap(listener);
```
AndyTWF, line 196: Nitpick: I think its cleaner to do something like:

`const wrapped = wrap(listener)` rather than reassigning

AndyTWF: (And in other places)

src/core/utils/event-emitter.ts
vladvelici: done

src/core/room-status.ts
vladvelici: done

AndyTWF (approved):

vladvelici:
@splindsay-92 I've added tests and found some listeners I forgot to wrap initially.

splindsay-92 (approved):
Nice, thanks for making the changes!

PR #524, opened 2025-04-14
Typing: Refactor typing events to use `SetChanged` type.

splindsay-92:
### Context

To provide better consistency with other events in Chat, the typing event payload now contains a `type` field at the root level. This also means `TypingEvents.Start` and `TypingEvents.Stop` have been replaced with `TypingEvents.SetChanged` for event emission.

### Checklist

* [x] QA'd by the author.

### DIFF
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -70,14 +70,9 @@ export enum PresenceEvents {
 }
 
 /**
- * All typing events.
+ * Enum representing the typing event types.
  */
-export enum TypingEvents {
-  /**
-   * Event triggered when a change occurs in the set of typers.
-   */
-  SetChanged = 'typing.set.changed',
-
+export enum TypingEventTypes {
   /**
    * Event triggered when a user is typing.
    */
@@ -90,13 +85,23 @@ export enum TypingEvents {
 }
 
 /**
+ * Enum representing the typing set event types.
+ */
+export enum TypingSetEventTypes {
+  /**
+   * Event triggered when a change occurs in the set of typers.
+   */
+  SetChanged = 'typing.set.changed',
+}
+
+/**
  * Represents a change in the state of current typers.
  */
-export interface TypingEvent {
+export interface TypingSetEvent {
   /**
    * The type of the event.
    */
-  type: TypingEvents;
+  type: TypingSetEventTypes;
 
   /**
    * The set of clientIds that are currently typing.
@@ -113,9 +118,9 @@ export interface TypingEvent {
     clientId: string;
 
     /**
-     * Type of the change. Either `typing.started` or `typing.stopped`.
+     * Type of the change.
      */
-    type: TypingEvents.Start | TypingEvents.Stop;
+    type: TypingEventTypes;
   };
 }
 
--- a/src/core/events.ts
+++ b/src/core/events.ts
@@ -74,6 +74,11 @@ export enum PresenceEvents {
  */
 export enum TypingEvents {
   /**
+   * Event triggered when a change occurs in the set of typers.
+   */
+  SetChanged = 'typing.set.changed',
+
+  /**
    * Event triggered when a user is typing.
    */
   Start = 'typing.started',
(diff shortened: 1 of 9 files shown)


src/core/events.ts
```diff
@@ -88,6 +93,11 @@ export enum TypingEvents {
  * Represents a change in the state of current typers.
  */
 export interface TypingEvent {
+  /**
+   * The type of the event.
+   */
+  type: TypingEvents;
```
sacOO7, line 99: This will always be `SetChanged`. Can we make it in a way that, it has two values,
`userSetUpdated`, `userSetCleared`

sacOO7: Also, this field will always have specific value that will not overlap with inner `type` field.
Do we still want to share common type for both?
IDE will give wrong cases when doing switch on `type` enum.
Maybe we should have separate types for both inner and outer fields.

sacOO7: Or just outer `type` field
```
export interface TypingEvent {

  type: TypingEvents; // alrady denotes what's changed

  clientId: string;

  currentlyTyping: Set<string>;
}


export enum TypingEvents {
  /**
   * Event triggered when a user is typing.
   */
  Start = 'typing.started',
  /**
   * Event triggered when a user stops typing.
   */
  Stop = 'typing.stopped',
}

```

sacOO7: Or if want to keep in sync with other events
```
export interface TypingEvent {

  type: TypingEvents; // alrady denotes what's changed

  data: {
    clientId: string;
    currentlyTyping: Set<string>;
  }
}


export enum TypingEvents {
  /**
   * Event triggered when a user is typing.
   */
  Start = 'typing.started',
  /**
   * Event triggered when a user stops typing.
   */
  Stop = 'typing.stopped',
}
```

splindsay-92: I can strongly type the change field like so..

```
/**
   * Represents the change that resulted in the new set of typers.
   */
  change: {
    /**
     * The client ID of the user who stopped/started typing.
     */
    clientId: string;

    /**
     * Type of the change. Either `typing.started` or `typing.stopped`.
     */
    type: TypingEvents.Start | TypingEvents.Stop
  };
```
wdyt?


> This will always be `SetChanged`. Can we make it in a way that, it has two values, `userSetUpdated`, `userSetCleared` or `clientTypingSetUpdated`, `clientTypingSetCleared`

We don't clear the set though, we only ever provide change to the set. So I'm not sure what benefit there is in adding an event that we don't use? 


> Also, this field will always have specific value that will not overlap with inner `type` field. Do we still want to share common type for both? IDE will give wrong cases when doing switch on `type` enum. Maybe we should have separate types for both inner and outer fields.

I had this chat with @AndyTWF this morning as I'd thought about splitting this out into separate enums, I still think there is merit there, but keeping all typing related events in the typing enum is not a bad choice. Plus, we may still want to expose individual Start/Stop events later.

splindsay-92: It is worth raising this in the DR though if you think the approach is wrong
   https://ably.atlassian.net/wiki/spaces/CHA/pages/3967713291/CHADR-101+Pre-GA+SDK+API+Review#Typing

sacOO7: Okay, I have added comment there.

AndyTWF: How about:

- `TypingEvent`, for changes related to an individual type (inner)
- `TypingSetEvent`, for changes related to the typing set itself (outer)

?

src/core/typing.ts
```diff
@@ -79,14 +79,13 @@ export interface Typing {
  * A listener which listens for typing events.
  * @param event The typing event.
  */
-export type TypingListener = (event: TypingEvent) => void;
+export type TypingListener = (event: TypingSetEvent) => void;
```
lawrence-forooghian, line 82: `TypingSetListener`?

lawrence-forooghian (approved):

PR #530, opened 2025-04-17
docs: avoid renaming functions in README example

owenpearson:
in my (highly subjective) opinion this looks a bit nicer - i think it looks a bit weird that the first example of code we have in the readme suggests overriding names of functions to avoid shadowing because it makes it look like our public APIs need to be overridden even for the simplest of use cases when it isn't actually necessary

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,8 +111,7 @@ async function getStartedWithChat() {
   // Create a new Ably Chat client, using the Ably client you created
   // The same client can be re-used for as long as your application is running
   const chatClient = new ChatClient(ablyClient);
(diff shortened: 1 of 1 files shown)


AndyTWF (approved):
I think this change is very reasonable - should we do the same in the docs repo too?

owenpearson:
@AndyTWF good shout, will do that now 👍

owenpearson:
actually - i don't think we use the same pattern anywhere in the docs repo so nothing to change there!

fwiw i think `const { off } = ...` which we use in all the docs examples is better than this, but if it causes a naming conflict then i prefer not destructuring at all

PR #535, opened 2025-05-02
Bugfix: empty summaries were dropped as invalid but in fact they are valid and mean there are no reactions (annotations)

vladvelici:
This happens when deleting the last reaction of a message.

### DIFF
--- a/src/core/messages-reactions.ts
+++ b/src/core/messages-reactions.ts
@@ -207,7 +207,9 @@ export class DefaultMessageReactions implements MessagesReactions {
       return;
     }
     if (!event.summary) {
-      return;
+      // This means the summary is now empty, which is valid.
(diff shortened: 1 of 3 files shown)


src/core/messages-reactions.ts
```diff
@@ -207,7 +207,9 @@ export class DefaultMessageReactions implements MessagesReactions {
       return;
     }
     if (!event.summary) {
-      return;
+      // This means the summary is now empty, which is valid.
+      // Happens when there are no reactions such as after deleting the last reaction.
+      event.summary = {};
```
AndyTWF, line 212: Is there a test case that covers this behaviour (I assume not?) - if not can we add one please

vladvelici: Added

AndyTWF (approved):

PR #551, opened 2025-05-20
Export MessageReactionEvents enum as value, not type

vladvelici:
### Context

I found this while working on adding message reactions the ably-cli. The enum is supposed to be exported as value, not as type.

### Checklist

* [x] QA'd by the author.


AndyTWF (approved):

owenpearson (approved):
looks good, i'm a bit hesitant about having typescript enums in our public API but that's a discussion for elsewhere 😄

PR #556, opened 2025-05-27
Fix flakey presence tests

splindsay-92:
### Context

[CHA-1003](https://ably.atlassian.net/browse/CHA-1003)

Presence tests were flakey, due to a race condition with sycn messages. A room is not subscribed on attach and the presence state is not in sync, if we then enter presence (and thus attach to the underlying channel) it is possible that the first protocol message we get will be a SYNC. If this is the case, then instead of an `enter` event being emitted back to our local listener, a `present` event is emitted. 

This isn't so much of a bug, as it is expected behaviour that we should account for in the tests.

### Description

* Presence integration tests now sync their state first by calling `presence.get()`, after which, `enter` events should be correctly received. 

### Checklist

* [x] QA'd by the author.

[CHA-101]: https://ably.atlassian.net/browse/CHA-101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CHA-1003]: https://ably.atlassian.net/browse/CHA-1003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/core/presence.integration.test.ts
+++ b/test/core/presence.integration.test.ts
@@ -91,6 +91,9 @@ describe('UserPresence', { timeout: 30000 }, () => {
     context.chat = newChatClient(undefined, context.realtime);
     context.defaultTestClientId = context.realtime.auth.clientId;
     context.chatRoom = await context.chat.rooms.get(roomId);
+
+    // Ensure we have just performed a sync so we don't get a `present` event instead of an `enter` event
+    await context.chatRoom.presence.get({ waitForSync: true });
   });
 
   // Test for successful entering with clientId and custom user data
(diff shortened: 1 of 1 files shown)


test/core/presence.integration.test.ts
```diff
@@ -91,6 +104,9 @@ describe('UserPresence', { timeout: 30000 }, () => {
     context.chat = newChatClient(undefined, context.realtime);
     context.defaultTestClientId = context.realtime.auth.clientId;
     context.chatRoom = await context.chat.rooms.get(roomId);
+
+    // Ensure we have just performed a sync so we don't get a `present` event instead of an `enter` event
+    await context.chatRoom.presence.get({ waitForSync: true });
```
AndyTWF, line 109: You'll want to do an `attach` call before this (will be checked after the V1 API review)

splindsay-92: Done

AndyTWF (approved):

PR #567, opened 2025-06-05
add reaction delete summary test

owenpearson:


test/core/messages-reactions.integration.test.ts
```diff
@@ -384,4 +384,126 @@ describe('message reactions integration', { timeout: 60000 }, () => {
...
+                total: 1,
+                clientIds: [client2.clientId],
+              },
+            },
+          },
+        });
+      },
+      { timeout: 30_000 },
+    );
+
```
vladvelici, line 464: Would it be a good idea to clear the `found` array here?

test/core/messages-reactions.integration.test.ts
```diff
@@ -384,4 +384,126 @@ describe('message reactions integration', { timeout: 60000 }, () => {
...
+              '❤️': {
+                total: 1,
+                clientIds: [client2.clientId],
+              },
+            },
+          },
+        });
+
+        // Check fourth summary (after deleting ❤️ - should be empty or not have these reactions)
+        const fourthSummary = found.findLast((e) => e.summary.messageSerial === message1.serial);
```
vladvelici, line 494: If the third one is not at index 2 (third in list), you'll find the third one here.

Would it be better to just findIndex the third one, and get the other index for the fourth one?

We're talking about indexes 2 and 3 for those checks, or 0 and 1 if you clear the found array.

I guess the same logic for finding the summaries in the list can be done for the first two as well?

vladvelici:
This will work now but I'm a bit worry about testing individual summaries after sending a few messages, because the idea with summaries is that we won't get one summary per annotation soon-ish (after we do rollups, or even now if we batch-publish the annotations).

PR #571, opened 2025-06-13
test coverage

AndyTWF:
### Description

* Turn off test coverage by default, as it slows things down quite a bit.
* Add test with coverage npm command
* Adds a number of tests in uncovered areas

### Checklist

* [x] QA'd by the author.

N/A

### DIFF
--- a/.cursor/rules/general.mdc
+++ b/.cursor/rules/general.mdc
@@ -13,4 +13,4 @@ If you see an opportunity for improvement, make a suggestion and ask for approva
 
 Update the tests whenever you make a change.
 
-Always run the tests.
+Always run the tests after making changes to source code or tests in order to catch any regressions.
--- a/.cursor/rules/general.mdc
+++ b/.cursor/rules/general.mdc
@@ -12,3 +12,5 @@ If you see an opportunity for improvement, make a suggestion and ask for approva
 # Always update the tests
 
 Update the tests whenever you make a change.
+
+Always run the tests.
--- a/src/core/rest-types.ts
+++ b/src/core/rest-types.ts
@@ -1,5 +1,5 @@
 import { ChatMessageAction } from './events.js';
-import { DefaultMessage, emptyMessageReactions, Message, MessageHeaders, MessageMetadata } from './message.js';
+import { DefaultMessage, emptyMessageReactions, Message } from './message.js';
 
 // RestClientIdList represents a list of client IDs with aggregation data
 export interface RestClientIdList {
@@ -35,7 +35,7 @@ export interface RestMessage {
   roomId: string;
   text: string;
   clientId: string;
-  action: string;
+  action: 'message.create' | 'message.update' | 'message.delete';
   metadata: Record<string, unknown>;
   headers: Record<string, string>;
   createdAt: number;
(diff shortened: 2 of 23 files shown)


src/core/rest-types.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/**
+ * Converts a message object from its REST representation to the standard message object in the SDK.
+ *
+ * @param message The message to convert from REST.
+ * @returns The converted message.
+ */
+export const messageFromRest = (message: RestMessage): Message => {
+  const metadata = message.metadata as MessageMetadata | undefined;
```
owenpearson, line 54: seems a bit sus to me that this function is supposed to generically convert a `RestMessage` to a `Message` but needs to make this additional assumption which isn't encapsulated by the `RestMessage` type. If this is a valid assumption then it should be part of that type, and if it's not then we can just assume the metadata + headers are defined so we wouldn't need these lines at all

src/core/rest-types.ts
```diff
@@ -0,0 +1,71 @@
...
+  const headers = message.headers as MessageHeaders | undefined;
+
+  const reactions = {
+    ...emptyMessageReactions(),
+    ...message.reactions,
+  };
+
+  return new DefaultMessage({
+    ...message,
+    action: message.action as ChatMessageAction,
```
owenpearson, line 64: same goes here - if it's already gonna be a `ChatMessageAction` then we should express that in the `RestMessage` type

src/core/version.ts
```diff
@@ -6,13 +6,13 @@ export const CHANNEL_OPTIONS_AGENT_STRING = `chat-js/${VERSION}`;
 export const CHANNEL_OPTIONS_AGENT_STRING_REACT = `chat-react/${VERSION}`;
 // Modes required for basic message functionality
 export const DEFAULT_CHANNEL_MODES: Ably.ChannelMode[] = ['PUBLISH', 'SUBSCRIBE'];
-export const DEFAULT_CHANNEL_OPTIONS: Ably.ChannelOptions = {
+export const DEFAULT_CHANNEL_OPTIONS = {
```
owenpearson, line 9: I don't see the advantage of this change, isn't it better to explicitly type this as a `ChannelOptions` object if that's how it's used?

owenpearson (changes requested):
new tests look good! few comments on the other changes:

.cursor/rules/general.mdc
```diff
@@ -11,4 +11,6 @@ If you see an opportunity for improvement, make a suggestion and ask for approva
 
 # Always update the tests
 
-Update the tests whenever you make a change.
\ No newline at end of file
+Update the tests whenever you make a change.
+
+Always run the tests.
```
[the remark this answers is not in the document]
AndyTWF: Done in [cb07c18](https://github.com/ably/ably-chat-js/pull/571/commits/cb07c1828b549922e56c90ff475b2faf9bbf5fbb)

src/core/rest-types.ts
AndyTWF: I've made the types better defined, and added a default in [1e0afc2](https://github.com/ably/ably-chat-js/pull/571/commits/1e0afc2db39681f711c7fc44d76911d267e228a7)

src/core/version.ts
AndyTWF: Fair, reverted back in [a9de20b](https://github.com/ably/ably-chat-js/pull/571/commits/a9de20b65a6791fec370becd522b2c0088b0baa4)

owenpearson (approved):
looks good!

PR #592, opened 2025-07-02
Add support for custom agent/version handling

splindsay-92:
### Context

* To allow us to gain insights into ui kit usage, we need a way to append the agent onto the connection.

### Description

* Added a new method `addAgentWithVersion` to the `ChatClient` class to allow for custom agent/version handling.
* ChatClientProvider now checks globalThis to see if an expected key exists for the ui kit.

### Checklist

* [x] QA'd by the author.

### DIFF
--- a/src/react/providers/chat-client-provider.tsx
+++ b/src/react/providers/chat-client-provider.tsx
@@ -35,15 +35,13 @@ export interface ChatClientProviderProps {
 export const ChatClientProvider = ({ children, client }: ChatClientProviderProps) => {
   const context = React.useContext(ChatClientContext);
   const value: ChatClientContextValue = React.useMemo(() => {
-    // Set the internal useReact option to true to enable React-specific agent.
-    (client as unknown as { addReactAgent(): void }).addReactAgent();
-
-    const uiKitVersion = (globalThis as Record<string, unknown>).__ABLY_CHAT_REACT_UI_COMPONENTS_VERSION__;
-    if (typeof uiKitVersion === 'string'){
-      (client as unknown as { addAgentWithVersion(agent: string, version:string): void }).addAgentWithVersion('chat-ui-kit', uiKitVersion);
+    client.addReactAgent();
+    const uiKitVersion = globalThis.__ABLY_CHAT_REACT_UI_COMPONENTS_VERSION__;
+    if (typeof uiKitVersion === 'string') {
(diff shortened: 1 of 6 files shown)


src/react/providers/chat-client-provider.tsx
```diff
@@ -37,6 +37,12 @@ export const ChatClientProvider = ({ children, client }: ChatClientProviderProps
   const value: ChatClientContextValue = React.useMemo(() => {
     // Set the internal useReact option to true to enable React-specific agent.
     (client as unknown as { addReactAgent(): void }).addReactAgent();
+
+    const uiKitVersion = (globalThis as Record<string, unknown>).__ABLY_CHAT_REACT_UI_COMPONENTS_VERSION__;
```
owenpearson, line 41: slightly cleaner to make a global.d.ts file with:
```typescript
declare global {
  interface Window {
    __ABLY_CHAT_REACT_UI_COMPONENTS_VERSION__?: string;
  }
}
```

owenpearson: you can use the same trick to declare the `addAgentWithVersion` method internally too :)

test/react/hooks/use-chat-client.test.tsx
```diff
@@ -50,6 +50,23 @@ describe('useChatClient', () => {
     render(<TestProvider />);
   });
 
+  it('should get the chat client from the context without error and ui kit agent if set', () => {
+    (globalThis as Record<string, unknown>).__ABLY_UI_KIT_VERSION__ = '1.0.0';
```
owenpearson, line 54: naming mismatch?
```suggestion
    (globalThis as Record<string, unknown>). __ABLY_CHAT_REACT_UI_COMPONENTS_VERSION__ = '1.0.0';
```

test/react/hooks/use-chat-client.test.tsx
```diff
@@ -50,6 +50,23 @@ describe('useChatClient', () => {
...
+    const chatClient = newChatClient();
+    const TestProvider = () => (
```
owenpearson, line 56: i think this test isn't extensive enough, is it possible to test that the agent actually gets sent over the wire. it's possible with this kind of set up that the client tries to connect before the window var is populated, in which case this test might pass but in reality the agent still doesn't get sent

owenpearson (approved):
lgtm 👍

PR #602, opened 2025-07-16
Fix: Handle presence re-entry fail.

splindsay-92:
### Context

* [CHA-1056](https://ably.atlassian.net/browse/CHA-1056)

The isPresent bool exposed by the `usePresence` hook doesn't take into account possible failures on auto re-entry. Also, the bool and error states could produce race conditions, since they are updated independently. 

### Description

To ensure the isPresent bool is accurate, we now handle auto re-entry failure that can occur and update the hook error state accordingly.
The isPresent and error states are now combined, to ensure no race conditions can occur in their rendering for users of the hook.

### Checklist

* [x] QA'd by the author.


src/react/hooks/use-presence.ts
```diff
@@ -61,7 +66,8 @@ const INACTIVE_CONNECTION_STATES = new Set<ConnectionStatus>([ConnectionStatus.S
  * A hook that provides access to the {@link Presence} instance in the room.
  * It will use the instance belonging to the room in the nearest {@link ChatRoomProvider} in the component tree.
  * On calling, the hook will `enter` the room with the provided data and `leave` the room when the component unmounts.
- * The {@link UsePresenceResponse.isPresent} flag will indicate when the user has become present in the room.
+ * The {@link UsePresenceResponse.presenceState} can be used to determine if the user is currently present in the room, and if any errors occurred while trying to enter or leave presence.
```
AndyTWF, line 69: I think we should add a comment here saying that if you get an error with code 91004, how you should handle it.

Also we could add 91004 to the `errors.ts` enum

src/react/hooks/use-presence.ts
```diff
@@ -1,4 +1,4 @@
-import { type ErrorInfo } from 'ably';
+import { ChannelStateChange, type ErrorInfo } from 'ably';
```
AndyTWF, line 1: Everywhere else we do `import * as Ably from 'ably'` so it's clear what is coming from ably-js and what isn't. Can we do the same here?

src/react/hooks/use-presence.ts
```diff
@@ -42,14 +42,19 @@ export interface UsePresenceResponse extends ChatStatusResponse {
...
-   * Indicates whether the current user is present in the room.
+   * The current presence state of this client.
    */
-  readonly isPresent: boolean;
-
-  /**
-   * Indicates if an error occurred while entering or leaving the room.
-   */
-  readonly error?: ErrorInfo;
+  readonly presenceState: {
```
AndyTWF, line 47: Is `presenceState` too generic? Should it be something like `userPresenceState` or `userState` ?

splindsay-92: I went with presenceState (or status could work) because we combine the error in there, so it felt better being a bit more generic. That said, I'd be happy with `userPresenceState`, or perhaps `myPresenceState`?

src/react/hooks/use-presence.ts
splindsay-92: At the moment, the advice would be to re-mount the hook, another reason why I think [making the mount more controllable](https://ably.atlassian.net/wiki/spaces/CHA/pages/4171857998/CHADR-114+Make+UsePresence+mount+behaviour+more+flexible) will be useful!

AndyTWF (commented):
Just the one response re whether it's `userPresenceState` or `myPresenceState` :)

src/react/hooks/use-presence.ts
```diff
@@ -170,27 +182,47 @@ export const usePresence = (params?: UsePresenceParams): UsePresenceResponse =>
...
+
+        return () => {
+          logger.debug('usePresence(); unsubscribing from presence state changes');
+          subscription.unsubscribe();
+        };
+      },
+      logger,
+    ).unmount();
+  }, [context, logger]);
+
```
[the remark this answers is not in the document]
splindsay-92: No, I don't want the onPresenceStateChange to be public as we may change both the API and internal logic.

src/core/presence.ts
```diff
@@ -320,4 +387,39 @@ export class DefaultPresence implements Presence {
...
+    const wrapped = wrap(listener);
+    this._stateEmitter.on('presence.state.change', wrapped);
+    return {
+      unsubscribe: () => {
+        this._logger.trace('Presence.unsubscribeFromPresenceStateChanges()');
+        this._stateEmitter.off(wrapped);
+      },
+    };
+  }
 }
```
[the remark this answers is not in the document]
splindsay-92: I don't want to export an internal interface, I have defined an internal interface in the hook where it's needed.

owenpearson (approved):
lgtm 👍

test/core/presence.test.ts
```diff
@@ -3,20 +3,33 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
...
 import { PresenceEventType } from '../../src/core/events.ts';
-import { DefaultPresence, PresenceData, PresenceEvent } from '../../src/core/presence.ts';
+import {
+  DefaultPresence,
+  Presence,
+  PresenceData,
+  PresenceEvent,
+  PresenceStateChange,
+  PresenceStateChangeListener,
+} from '../../src/core/presence.ts';
```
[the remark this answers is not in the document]
splindsay-92: There is no need, they are not part of the public API since they are not included in the index.ts file exports. While a user could access them by importing the file directly, it is clear they are not part of the public API and so we make no commitment not to change the function signature (or remove them entirely) at any time.

PR #655, opened 2025-10-03
chat-client: make clientId optional value

AndyTWF:
### Context

[CHA-1195]

### Description

ably-js has incorrect typings, but the auth.clientId is undefined in the
case that token auth is used and we have not yet connected to the server
(RSA12b).

This change updates the chat-client and useChatClient hook to make
clientId an optional value, to match the spec.

Also fixed:

- Updated the realtime client with token test helper to delete the sandbox key and clientId from the options, to get "true token auth"
- Added a helper for tests that need a client ID to wait until the ChatClient has one.

### Checklist

* [x] QA'd by the author.

N/A


src/core/client-id.ts
```diff
@@ -0,0 +1,46 @@
...
+    this._clientId = realtime.auth.clientId;
+    this._off = connection.onStatusChange((event: ConnectionStatusChange) => {
+      if (event.current === ConnectionStatus.Connected) {
+        this._clientId = realtime.auth.clientId;
+        this._logger.debug('resolved clientId', { clientId: this._clientId });
+      }
+    }).off;
+  }
+
+  get(): string {
```
ttypic, line 35: Should we really subscribe on status change? can we simply 

```ts
get(): string {
   const clientId = realtime.auth.clientId;
   if (!clientId) {
      throw new Ably.ErrorInfo('invalid client id', 40012, 400);
    }

    return clientId;
}
```

src/core/messages.ts
```diff
@@ -542,7 +543,7 @@ export class DefaultMessages implements Messages {
     const timestamp = new Date(response.timestamp);
     return new DefaultMessage({
       serial: response.serial,
-      clientId: this._clientId,
+      clientId: this._clientId.get(),
```
lawrence-forooghian, line 546: You're allowed to send a message without establishing a connection, right? So this might unnecessarily fail given that the resolver depends the establishment of a realtime connection.

By the time that the REST API call has returned, the core SDK should have fetched a token (since it needs one to make the REST request), so I think that if you change the resolver to not wait for a connection and instead just always re-fetch the client ID from the realtime client then it should work.

(will look for a core SDK spec point that confirms this)

AndyTWF: I've actually got a separate PR to get this information from the server response directly, so this will go away shortly anyway!

AndyTWF: So interestingly, this is a problem for JWTs. If you're using a JWT then you have two options:

- Force a `clientId` in `ClientOptions`
- Connect to Ably first

As `TokenRequest` and `TokenDetails` aren't used in that flow

